PR #117, opened 2021-03-13
change the styles of coordinator header buttons

Marwanrizik91:
- match the design
- changed start script in package.json in coordinator(was not running for me(windows))

image

### DIFF
--- a/packages/coordinator/src/components/Header/CoordinatorHeader.tsx
+++ b/packages/coordinator/src/components/Header/CoordinatorHeader.tsx
@@ -1,5 +1,5 @@
 import Button from "../Button";
-import React from "react";
+import React, { useState } from "react";
 import logoImg from "../../assets/logo.svg";
 import styles from "./CoordinatorHeader.module.scss";
 import { useHistory } from "react-router-dom";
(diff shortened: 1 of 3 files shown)


packages/coordinator/src/components/Header/CoordinatorHeader.tsx
```diff
@@ -13,23 +13,30 @@ interface CoordinatorHeaderProps {
 export default function CoordinatorHeader(props: CoordinatorHeaderProps) {
   const history = useHistory();
 
-  const navigate = (screen: CoordinatorScreen) => () =>
+  const [currentTab, setCurrentTab] = useState("home");
+
+  const navigate = (screen: CoordinatorScreen) => () => {
```
yaron1m, line 18: The problem with this is that it won't change when users hit "back" or manually navigate to a different screen.

Is there a way to use the history hook to know what page we are in?

Marwanrizik91: fixing it, initially started doing it with useLocation, then changed to a state.
will update soon

yaron1m (approved):

PR #121, opened 2021-03-18
Add minimum timeout to splash screen

yaron1m:
Adi and Merav asked that the splash screen would have a minimum appearance time.


yaron1m:
@Adi-q Please advise what should be the minimum time

Adi-q:
> @Adi-q Please advise what should be the minimum time

2 seconds.

evbambly (approved):

PR #122, opened 2021-03-18
Donors table

yaron1m:
image


packages/common/src/SortingUtils.ts
```diff
@@ -0,0 +1,3 @@
+export function StringComparator<T>(fieldSelector: (x: T) => String) {
+  return (a: T, b: T) => (fieldSelector(a) > fieldSelector(b) ? 1 : -1);
+}
```
evbambly, line 3: Should the Sort function be joined to this file?

How should we organize utils / function types in our code?

evbambly (approved):

packages/common/src/SortingUtils.ts
yaron1m: Will move the other one to common as well

PR #124, opened 2021-03-21
Revert "Revert "questionnaire wrong answer popup""

harelglik:
This reverts commit e29c23aa81969d10e5fde69f91e95821277cc751.

### DIFF
--- a/packages/donor/src/components/basic/Popup/Popup.tsx
+++ b/packages/donor/src/components/basic/Popup/Popup.tsx
@@ -12,7 +12,7 @@ type PopupProps = {
   titleSecond?: string;
   className?: string;
   isLoading?: boolean;
-  onClose?: () => void;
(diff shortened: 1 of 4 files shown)


packages/donor/src/components/basic/Popup/Popup.tsx
```diff
@@ -47,14 +48,16 @@ export default function Popup({
...
-        <hr className={styles.hr}></hr>
+        <hr className={styles.header}></hr>
         <DialogActions>
-          <Button
-            onClick={onClose}
-            title="חזרה"
-            className={styles.backButton}
-            variant={ButtonVariant.text}
-          />
+          {onClose && (
```
yaron1m, line 53: I think this should be onBack

yaron1m (approved):

PR #129, opened 2021-03-22
Coord table fixes

evbambly:
image

* Changed default sorting method to default sorting column index
* Added indications of which column is currently sorting, as well as a cursor to sortable columns

### DIFF
--- a/packages/coordinator/src/screens/serchDonorsScreen/SearchDonorsScreen.tsx
+++ b/packages/coordinator/src/screens/serchDonorsScreen/SearchDonorsScreen.tsx
@@ -53,9 +53,7 @@ export default function SearchDonorsScreen({
         hasColumnHeaders
         columns={columns}
         rows={rows}
-        initialRowSorter={SortingUtils.StringComparator<Donor>(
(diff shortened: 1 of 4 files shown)


packages/coordinator/src/screens/serchDonorsScreen/SearchDonorsScreen.tsx
```diff
@@ -53,9 +53,7 @@ export default function SearchDonorsScreen({
         hasColumnHeaders
         columns={columns}
         rows={rows}
-        initialRowSorter={SortingUtils.StringComparator<Donor>(
-          (d) => d.firstName
-        )}
+        initialSortByColumnIndex={0}
```
yaron1m, line 56: Is this really required?

yaron1m (approved):

packages/coordinator/src/screens/serchDonorsScreen/SearchDonorsScreen.tsx
evbambly: Nope!

PR #133, opened 2021-03-23
Coordinator app version

yaron1m:
To be honest, I am not sure about the CSS I added, please verify it makes sense

image

### DIFF
--- a/packages/coordinator/src/navigation/CoordinatorRouter.module.scss
+++ b/packages/coordinator/src/navigation/CoordinatorRouter.module.scss
@@ -7,8 +7,7 @@ $footer_height: var(--height-footer, 30px);
 .footer {
   color: var(--color-mid-grey);
   height: $footer_height;
-  position: absolute;
-  bottom: 0;
-  left: 2px;
   font-size: 14px;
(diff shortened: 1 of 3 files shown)


packages/coordinator/src/navigation/CoordinatorRouter.module.scss
```diff
@@ -1,3 +1,14 @@
...
+$footer_height: var(--height-footer, 30px);
+
+.footer {
+  color: var(--color-mid-grey);
+  height: $footer_height;
+  position: absolute;
+  bottom: 0;
+  left: 2px;
+  font-size: 14px;
 }
```
evbambly, line 14: 1. Since this div is positioned inside and with the flow, no reason to use absolute positioning (or any explicit position)
2. This part of the website is written in English only and can use the "direction: ltr" attribute to position it to the left

evbambly: 3. Naturally, absolute positioning related attributes (bottom, left) can be replaced with relative attributes (padding, margin)

evbambly (approved):

PR #136, opened 2021-03-26
Remove donor button

yaron1m:
* Add new button
* Add Tooltip to button
* Update BE function

image

image

image

### DIFF
--- a/packages/coordinator/src/components/IconButton/IconButton.tsx
+++ b/packages/coordinator/src/components/IconButton/IconButton.tsx
@@ -5,6 +5,7 @@ import {
   SvgIconTypeMap,
 } from "@material-ui/core";
 import { OverridableComponent } from "@material-ui/core/OverridableComponent";
+import Tooltip from "@material-ui/core/Tooltip";
 
 export enum IconSize {
   small = "small",
@@ -13,6 +14,7 @@ export enum IconSize {
 
 export enum Icon {
   Delete = "Delete",
+  Clear = "Clear",
 }
 
 export interface IconButtonProps {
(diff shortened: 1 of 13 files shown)


packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsScreenContainer.tsx
```diff
@@ -39,6 +40,19 @@ export default function ManageAppointmentsScreenContainer() {
...
   };
 
+  const onRemoveDonor = (appointmentId: string) => {
+    setAppointmentsResponse({
+      ...appointmentsResponse,
+      appointments: appointmentsResponse.appointments.map((appointment) => {
+        if (appointment.id !== appointmentId) {
+          return appointment;
+        }
+        return AppointmentUtils.removeDonorFromAppointment(appointment);
```
evbambly, line 50: Is the process removing the donor twice,
once for the frontend and one via Firebase?

Why not amend the appointments and load them again- keeping a single source of truth?

packages/coordinator/src/components/IconButton/IconButton.tsx
```diff
@@ -5,6 +5,7 @@ import {
   SvgIconTypeMap,
 } from "@material-ui/core";
 import { OverridableComponent } from "@material-ui/core/OverridableComponent";
+import Tooltip from "@material-ui/core/Tooltip";
```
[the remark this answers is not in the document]
yaron1m: Do you think we need it anywhere else besides buttons for now?

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsScreenContainer.tsx
yaron1m: Well, we could, but that means higher latency. Why is it butter, when both BE and FE use the same util?

evbambly (approved):

PR #147, opened 2021-04-17
burger menu button

harelglik:
Kapture 2021-04-18 at 01 55 11

### DIFF
--- a/packages/donor/src/components/AppHeader/AppHeader.tsx
+++ b/packages/donor/src/components/AppHeader/AppHeader.tsx
@@ -10,6 +10,11 @@ import { ArrowForward } from "@material-ui/icons";
 import { useHistory } from "react-router-dom";
 import { MainNavigationKeys } from "../../navigation/app/MainNavigationKeys";
 import profileIcon from "../../assets/icons/profile.svg";
+import { BurgerMenu } from "./BurgerMenu";
+import { useState } from "react";
+import { SideDrawer } from "./BurgerMenu/SideDrawer";
+import { Backdrop } from "../Backdrop";
+import firebase from "firebase";
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
@@ -28,15 +33,18 @@ interface AppHeaderProps {
   title: string;
   hasBackButton?: boolean;
   hasProfileButton?: boolean;
+  hasBurgerMenu?: boolean;
 }
 
 export default function ButtonAppBar({
   hasBackButton,
   hasProfileButton,
   title,
+  hasBurgerMenu,
 }: AppHeaderProps) {
   const classes = useStyles();
   const history = useHistory();
+  const [showSideBar, setShowSideBar] = useState(false);
 
   if (!title) {
     console.error("Unknown pathname");
@@ -45,6 +53,13 @@ export default function ButtonAppBar({
     <AppBar position="fixed" color="secondary">
       <Toolbar>
         <Grid container>
+          <Grid item xs>
+            {hasBurgerMenu && (
+              <BurgerMenu
+                onClick={() => setShowSideBar((previous) => !previous)}
+              />
+            )}
+          </Grid>
           <Grid item xs>
             {hasBackButton && (
               <IconButton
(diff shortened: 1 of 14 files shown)


packages/donor/src/components/AppHeader/AppHeader.tsx
```diff
@@ -78,6 +93,25 @@ export default function ButtonAppBar({
...
+          close={() => setShowSideBar(false)}
+          navProps={{
+            navItemsProps: [
+              {
+                title: "החשבון שלי",
+                onClick: () => history.push("/" + MainNavigationKeys.MyProfile),
+              },
+              {
+                title: "התנתק",
+                onClick: () => firebase.auth().signOut(),
```
yaron1m, line 108: 1. Take signOut() as a prop here
2. Remove the sign out option from the profile page

packages/donor/src/components/AppHeader/AppHeader.tsx
```diff
@@ -28,15 +33,18 @@ interface AppHeaderProps {
   title: string;
   hasBackButton?: boolean;
   hasProfileButton?: boolean;
+  hasBurgerMenu?: boolean;
```
yaron1m, line 36: Do you want to add a new story to the storybook?

packages/donor/src/components/AppHeader/AppHeader.tsx
```diff
@@ -45,6 +53,13 @@ export default function ButtonAppBar({
     <AppBar position="fixed" color="secondary">
       <Toolbar>
         <Grid container>
+          <Grid item xs>
+            {hasBurgerMenu && (
+              <BurgerMenu
+                onClick={() => setShowSideBar((previous) => !previous)}
```
yaron1m, line 59: isnt this enough?
`onClick={() => setShowSideBar(!showSideBar)}`

packages/donor/src/components/AppHeader/AppHeader.tsx
```diff
@@ -78,6 +93,25 @@ export default function ButtonAppBar({
           </Grid>
         </Grid>
       </Toolbar>
+      <div>
+        <SideDrawer
+          isOpen={showSideBar}
+          close={() => setShowSideBar(false)}
+          navProps={{
+            navItemsProps: [
```
yaron1m, line 101: Please add the about page (here or in another PR)

packages/donor/src/components/basic/ZMScreen/ZMScreen.tsx
```diff
@@ -6,20 +6,20 @@ interface ExtendedSignupScreenProps {
...
   className = "",
   hasBackButton,
-  hasProfileButton,
+  hasBurgerMenu,
   title,
 }: ExtendedSignupScreenProps) {
   return (
     <>
-      <AppHeader {...{ hasBackButton, hasProfileButton, title }} />
+      <AppHeader {...{ hasBurgerMenu, hasBackButton, title }} />
```
yaron1m, line 22: Should we somehow enforce we don't get both back button and burger menu?

packages/donor/src/components/AppHeader/AppHeader.tsx
```diff
@@ -78,6 +93,25 @@ export default function ButtonAppBar({
...
+              {
+                title: "החשבון שלי",
```
yaron1m, line 103: The page title says פרופיל so let's be consistent

packages/donor/src/components/AppHeader/AppHeader.tsx
harelglik: The correct way to update a state based on the previous state in react is to use the overloaded that gives you the previous state and not rely on the state value.

packages/donor/src/components/basic/ZMScreen/ZMScreen.tsx
harelglik: Don't think we should as in theory there is no reason why other pages should not be able to show both.

packages/donor/src/components/AppHeader/AppHeader.tsx
harelglik: הפרופיל שלי

packages/donor/src/components/AppHeader/AppHeader.tsx
harelglik: done

packages/donor/src/components/AppHeader/AppHeader.tsx
harelglik: I started with the signout coming from outside but then you have to specify it and it's not always used to it becomes a little clumsy.
Is there a specific reason (testing?) why it's problematic to have it inlined here?

packages/donor/src/components/AppHeader/AppHeader.tsx
harelglik: It's a bit trickier with the popups but I will try

packages/donor/src/components/AppHeader/AppHeader.tsx
harelglik: 2 = done

yaron1m (approved):

PR #148, opened 2021-04-20
button v2

talwag:
Screenshot 2021-04-20 185908
Screenshot 2021-04-20 185843
Screenshot 2021-04-20 185759
Screenshot 2021-04-20 185823


packages/donor/src/components/basicV2/Button/Button.stories.tsx
```diff
@@ -0,0 +1,101 @@
+import Button, { ButtonProps } from "./Button";
+import { MuiTestWrapper } from "../../../__test__/TestUtils";
+
+export default {
+  component: Button,
+  title: "Buttonv2",
```
yaron1m, line 6: I changed the convention of the titles, please use `Components V2/Button`

packages/donor/src/components/basicV2/Button/Button.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+const baseArgs = {
+  title: "אישור והמשך",
+  onClick: () => {},
+};
+
+export const Main = (args: ButtonProps) => (
+  <MuiTestWrapper>
+    <Button {...args} />
+  </MuiTestWrapper>
```
yaron1m, line 17: I changed this in master so that there's no more need for code duplication, or the test wrapper.

Please merge from master and see how Button.stories looks now, and do the same here.

(My bad for only doing this now, sorry)

yaron1m (approved):

PR #154, opened 2021-04-28
Add capacitor support 

MorisR:
### Added Global Dependencies:
```inquirer, yargs``` - Used to run "build-dev-apk" script with a "cli' interface.
```fs-extra``` - Expands upon "fs" built in package.
```npm-run-all``` - Allows to run multiple npm scripts at once in parallel or in sequence.
```ts-node``` - Allow running typescript files via node directly w/o having to transpile them to js first.

### Donor Dependencies:
```@capacitor/*``` - Native web view wrapper dependencies.
```cordova-res``` - Allows generating logo + splash screen images for native apps.

### Donor Scripts:
```mobile:init-ios``` - Generates ios empty/initial web view wrapper.
```mobile:init-android``` - Generate android empty/initial web view wrapper.
```mobile:update-deps``` - Updates the dependencies of the added wrappers.
```mobile:generate-images``` - Generates logo and splash for both android and ios
```mobile:start-dev``` or  ```start-mobile```- Runs the in mobile simulator in development mode including hot reload, (```start``` script must be running in the background).
```sync-mobile``` - copies all the changes from the "build" directory to the mobile wrappers. (must run ```build``` first)
```build-dev-apk``` - Generates and sign the apk with a demi/empty certificate and places it in the donor root dir
 
 
### Changelog:
- Allow wapping the donor app in mobile views for both android and ios.
- Add global css classes ```safe-screen-padding-top``` , ```safe-screen-padding-bottom```, ```safe-screen-padding-left``` and ```safe-screen-padding-right``` that applies padding to tackle the notch issue.
- Add ```SafeScreen``` and ```SafeArea``` that applies the same padding.
- Link js colors with css colors so that we won't have to modify them twice each time.

### Notes
- Here are the pre-requirements for capacitor to work locally https://capacitorjs.com/docs/getting-started/dependencies
- Usually the process of building everything requires running the following commands in sequence 
(if the mobile apps weren't generated, you need to run ```init-mobile:ios``` and ```init-mobile:android```)
```build``` --> ```sync-mobile``` --> ```generate-mobile-images```

### Suggestions
- I can remove the ```build-dev-apk``` command and along side it also remove ```inquirer``` and  ```yargs``` deps.
- We should add the "ios" and "android" build directories to the repo in a separate branch along side github action to validate the scripts.
- Maybe we should move all the ```capacitor``` deps to the global/base ```package.json``` file.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -36,6 +36,7 @@
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/pickers": "^3.2.10",
+    "@types/fs-extra": "^9.0.11",
```
yaron1m, line 39: Should this be a dev-dep?

package.json
```diff
@@ -58,18 +59,25 @@
     "@storybook/node-logger": "^6.1.21",
```
yaron1m, line 59: Can we remove `react-div-100vh`?

packages/donor/src/screens/HomeScreen.tsx
```diff
@@ -1,5 +1,6 @@
 import Button from "../components/basic/Button";
 import Text from "../components/basic/Text";
+import SafeScreen from "../components/basic/SafeScreen";
```
yaron1m, line 3: What does this do?

packages/donor/tsconfig.json
```diff
@@ -27,5 +27,12 @@
   },
   "include": [
     "src"
-  ]
+  ],
+  "ts-node": {
```
yaron1m, line 31: Should this also be in `tsconfig.build.json`?

packages/donor/src/screens/HomeScreen.tsx
MorisR: It adds padding to the page to accommodate for the notch + the lower bar on iphones , you can also disable that padding via the props if need be.

package.json
MorisR: The "react-div-100vh" was necessary because we released it as a website, and it's used to accommodate for the browser status bar height changes.
Now that we're releasing an app, it's not necessary anymore.

We just need to make sure it's not used in the coordinator app

package.json
MorisR: My bad, yeah

packages/donor/src/screens/HomeScreen.tsx
MorisR: Same as the "safe area" component, just with full width and height

packages/donor/tsconfig.json
MorisR: I'll get back to you on this one,
Not really sure,
I manly use it to generate the signed apk file,
( Running npm scripts with ts)

Which is a dev script? 🤷🏻‍♂️

MorisR: it's ignoring the "extends" property.

that's the current config inheritance "tree":
```/tsconfig.build.json``` --> ```/tsconfig.json``` --> ```packages/donor/tsconfig.json```

I've tried moving the "ts-node" field to the other two files with no luck, the scripts that depend on it are failing.

package.json
yaron1m: I think we are still going to release this app in web in the coming weeks.

In such cases, should we leave it for now?

packages/donor/tsconfig.json
yaron1m: I am not super familiar with this, so I trust you.

If you need help here, contact Harel

package.json
MorisR: I'll re-add it then

yaron1m (approved):

PR #157, opened 2021-04-30
Feature/new scheduled booking

evbambly:
image
image

New Bookings are now shown as new.
If Only New Bookings button is selected, 
new bookings/cancellations are shown without a tag.


packages/coordinator/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,103 @@
...
+  label: string;
+  /* Standard template chips **/
+  chipType?: StandardChip;
+  color?: string;
+  backgroundColor?: string;
+  fontSize?: string;
+  /* Applying width disables default padding on label span **/
+  width?: string;
+  height?: string;
+};
```
yaron1m, line 82: 2 comments:
1. If we have `chipType` why do we also need `label`?
2. I don't like doing הכנה למזגן. If currently no one needs custom height and width, I don't think we should support such props. When we need them, we can add them

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsScreen.tsx
```diff
@@ -32,6 +32,7 @@ interface ManageAppointmentsScreenProps {
   onDeleteAppointment: (appointmentId: string) => Promise<void>;
   onRemoveDonor: (appointmentId: string) => Promise<void>;
   isLoading: boolean;
+  hasOnlyNewChanges: boolean;
```
yaron1m, line 35: showOnlyWithRecentChanges

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsScreenContainer.tsx
```diff
@@ -20,6 +20,7 @@ export default function ManageAppointmentsScreenContainer() {
     getDefaultState()
   );
   const [isLoading, setIsLoading] = useState(false);
+  const [showOnlyNewChanges, setShowOnlyNewChanges] = useState(false);
```
yaron1m, line 23: Please go over the PR and be consistent with the naming. 
Now we have `showOnlyNew`, `ChangeType`, `BookingChange` and they all refer to the same thing

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
```diff
@@ -29,11 +31,35 @@ export const GetExpandedColumns = (
...
+    cellRenderer: ({ booked, bookingTimeMillis, recentChangeType }) => {
+      let bookingDate = "אין רישום";
+      if (booked && bookingTimeMillis) {
+        bookingDate =
+          "נקבע בתאריך " + DateUtils.ToDateString(bookingTimeMillis);
       }
-      return "נקבע בתאריך " + DateUtils.ToDateString(bookingTimeMillis);
+      const chipLabel =
+        recentChangeType === BookingChange.CANCELLED ? "בוטל" : "נקבע";
+      return (
```
yaron1m, line 42: Should this just be a new column?

This way we could possibly even sort by it.

WDYT?

packages/coordinator/src/components/Chip/Chip.tsx
evbambly: ChipType's name is now ColorScheme for clarity.

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsScreenContainer.tsx
evbambly: showOnlyNew converted to showOnlyRecentChanges,
in consistency with RecentChangeType.

BookingChange refers to both recent and non-recent changes

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
evbambly: I agree, it might be more convenient

yaron1m (approved):

PR #161, opened 2021-05-06
HTML notification emails

yaron1m:


packages/functions/src/dal/EmailNotificationsDataAccessLayer.ts
```diff
@@ -2,10 +2,10 @@ import * as admin from "firebase-admin";
 import { Collections } from "@zm-blood-components/common";
 
 export type EmailMessage = {
-  to: string[];
+  to: string | string[];
```
harelglik, line 5: Why more than 1 type?

harelglik (approved):

packages/functions/src/dal/EmailNotificationsDataAccessLayer.ts
yaron1m: This is what the framework support, either one email or an array of emails. both work

PR #165, opened 2021-05-12
groups

harelglik:


packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.test.ts
```diff
@@ -41,56 +37,46 @@ const ALL_APPOINTMENT_IDS = [
...
 beforeAll(reset);
 afterEach(reset);
-
 test("Unauthenticated user throws exception", async () => {
   const action = () => callFunction();
   await expectAsyncThrows(action, "Unauthorized");
 });
 
-test("User that is not admin throws exception", async () => {
+test("User that is not coordinator throws exception", async () => {
```
yaron1m, line 53: Should you add more tests to this file? GetCoordinatorAppointmentsHandler changed a lot

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
```diff
@@ -28,11 +31,69 @@ export default async function (
...
+    case CoordinatorRole.GROUP_COORDINATOR:
+      break;
+  }
+}
+
+async function filterDonorsInGroup(
+  coordinator: DbCoordinator,
+  donorsInAppointments: DbDonor[]
+) {
+  const groupIds = await GroupDAL.getGroupIdsOfCoordinatorId(coordinator.id);
```
yaron1m, line 85: Should this be Set?

packages/functions/src/coordinator/GetDonorsHandler.ts
```diff
@@ -1,18 +1,40 @@
...
+  const coordinator = await CoordinatorDAL.getCoordinator(callerId);
   if (!coordinator) {
     console.error("Could not find calling user", callerId);
     throw Error(`User ${callerId} is not an admin`);
   }
 
-  const donors = await getAllDonors();
+  let donors: DbDonor[] = [];
+
+  switch (coordinator.role) {
```
yaron1m, line 22: Tests?

packages/functions/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
yaron1m, line 1: Remove this file

packages/functions/tsconfig.json
```diff
@@ -1,3 +1,24 @@
 {
-  "extends": "../../tsconfig.json"
+  "extends": "../../tsconfig.json",
```
yaron1m, line 2: Revert this change

packages/functions/src/react-app-env.d.ts
harelglik: Done

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.test.ts
harelglik: Sorry did not merge the test files correctly, there is now a test that checks donors in groups

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
harelglik: Done
There are many places where we should use sets but use arrays.

yaron1m (approved):

PR #171, opened 2021-05-22
Feature/reports frontend

evbambly:
image

Export BookedDonationWithDonorDetails[] to CSV


packages/common/src/LocaleUtils.ts
```diff
@@ -35,7 +35,7 @@ export const getBloodTypeTranslation = (bloodType: BloodType) => {
     case BloodType.O_PLUS:
       return "O+";
     case BloodType.NOT_SURE:
-      return "לא יודע";
+      return "לא יודע/ת";
```
yaron1m, line 38: Didn't you just change this line in a different PR?

packages/coordinator/src/components/CsvDownloaderButton/CsvDownloaderButton.tsx
```diff
@@ -0,0 +1,41 @@
+import CsvDownloader from "react-csv-downloader";
+import { IColumn } from "react-csv-downloader/dist/esm/lib/csv";
+import Button from "../Button";
+
+export interface ZMCsvDownloaderButtonProps {
+  columns?: IColumn[];
+  data: { [key: string]: any }[];
+  fileName: string;
+  separator?: string;
```
yaron1m, line 9: In my opinion, if no one uses a different value than `,` we should not expose it.

We can always add this when needed

packages/coordinator/src/components/CsvDownloaderButton/CsvDownloaderButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  title: string;
+  onClick?: () => void;
```
yaron1m, line 11: With respect to the previous comment, I believe when a new component like this is added it should not have optional values (especially when there's currently only 1 caller).

Either a prop is required, or it should not exist for now

packages/coordinator/src/components/Header/CoordinatorHeader.tsx
```diff
@@ -48,6 +48,11 @@ export default function CoordinatorHeader(props: CoordinatorHeaderProps) {
           title="חיפוש משתמשים"
           onClick={navigate(CoordinatorScreen.DONORS)}
         />
+        <Button
+          className={currentTab === "/donors" ? styles.selected__tab : ""}
```
yaron1m, line 52: `currentTab === "/booked-donations"`

packages/coordinator/src/screens/scheduledAppointments/ScheduledAppointmentsScreenContainer.tsx
```diff
@@ -0,0 +1,95 @@
...
+      await CoordinatorFunctions.getBookedAppointmentsInHospital({
+        hospital,
+        fromDateMillis: fromDateMidnight!.getTime(),
+        toDateMillis: toDate!.getTime(),
+      });
+    setAppointmentsWithDonorDetails(nextAppointmentsWithDonorDetails);
+    setIsLoading(false);
+  };
+
+  return (
```
yaron1m, line 49: The `return` of a container should have a single React element (`ScheduledAppointmentsScreen`) and its props.

All other component should be part of the component itself and not the container.

We should also fix this for the other screens (but in a different PR please)

packages/coordinator/src/screens/scheduledAppointments/ScheduledAppointmentsScreenContainer.tsx
```diff
@@ -0,0 +1,95 @@
...
+    <>
+      <HeaderSection className={styles.component}>
+        <Select
+          id={"hospital"}
+          label={"בית חולים"}
+          options={HospitalUtils.getAllHospitalOptions()}
```
yaron1m, line 55: I added an ability to show only hospitals relevant to coordinator role.

Please see implementation in other screens and do the same here

packages/coordinator/src/screens/scheduledAppointments/ScheduledAppointmentsScreenContainer.tsx
```diff
@@ -0,0 +1,95 @@
...
+  csvColumns,
+  formatDataByColumns,
+} from "./ScheduledAppointmentsCsvConfig";
+
+export default function ScheduledAppointmentsScreenContainer() {
+  const TWO_WEEKS_IN_MILLIS = 1000 * 60 * 60 * 24 * 14;
+  const initialToDate = new Date(Date.now() + TWO_WEEKS_IN_MILLIS);
+
+  const [fromDate, setFromDate] = useState<Date | null>(new Date());
+  const [toDate, setToDate] = useState<Date | null>(initialToDate);
```
yaron1m, line 26: Could it really be null?

packages/common/src/LocaleUtils.ts
evbambly: Will be solved in merge

package.json
```diff
@@ -44,6 +44,7 @@
     "lodash": "^4.17.20",
     "normalize.css": "^8.0.1",
     "react": "^17.0.1",
+    "react-csv-downloader": "^2.4.0",
```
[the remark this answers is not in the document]
evbambly: Will it install upon 'yarn install' of the root folder?

packages/coordinator/src/components/CsvDownloaderButton/CsvDownloaderButton.tsx
evbambly: onClick is required for the button component.
So why not expose it externally as well, rather than a stump*?
*onClick={()=>{}}

packages/coordinator/src/screens/scheduledAppointments/ScheduledAppointmentsScreenContainer.tsx
evbambly: Will be done in separate PR

package.json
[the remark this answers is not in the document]
yaron1m: Yes

yaron1m (approved):

PR #172, opened 2021-05-22
Bugfix/add blood type to manage donations

evbambly:
image


packages/coordinator/src/screens/manageAppointmentsScreen/CoordinatorAppointmentsGrouper.ts
```diff
@@ -24,6 +25,7 @@ export type ManagedAppointment = {
   bookingTimeMillis?: number;
   recentChangeType?: BookingChange;
   isPastAppointment: boolean;
+  bloodType: BloodType | undefined;
```
yaron1m, line 28: bloodType?: BloodType

yaron1m (approved):

PR #175, opened 2021-05-22
Increased table usibility

evbambly:
image

1. Make input form row into column
2. Make table width-scrollable like in the second example in the article*.
3. Minimum column separation (1rem)
4. Limit side scrolling on entire app (now width = 100%)


yaron1m (approved):

yaron1m:
It requires some work, but what do you think about adding storybook to this project and adding a story to the table?

This would make reviewing this UI change (for me and for Merav) much easier

PR #187, opened 2021-06-02
donor diconnect button down in side drawer

harelglik:
image


yaron1m:
Why isn't the white drawer all the way down?

PR #188, opened 2021-06-02
coordinator show only allowed pages

harelglik:
image
image
image
image

### DIFF
--- a/packages/coordinator/src/navigation/CoordinatorRouter.tsx
+++ b/packages/coordinator/src/navigation/CoordinatorRouter.tsx
@@ -4,7 +4,11 @@ import LoadingScreen from "../screens/loading/LoadingScreen";
 import { Redirect, Route, Switch } from "react-router-dom";
 import { CoordinatorScreen } from "./CoordinatorScreen";
 import AddAppointmentsScreenContainer from "../screens/addAppointments/AddAppointmentsScreenContainer";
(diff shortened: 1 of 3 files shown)


packages/coordinator/src/navigation/CoordinatorRouter.tsx
```diff
@@ -75,35 +122,49 @@ export default function CoordinatorRouter() {
...
-          <ScheduledAppointmentsContainer
-            activeHospitalsForCoordinator={activeHospitalsForCoordinator}
-          />
-        </Route>
-        <Route exact path={["/" + CoordinatorScreen.ADD_APPOINTMENTS, "*"]}>
-          <AddAppointmentsScreenContainer
-            activeHospitalsForCoordinator={activeHospitalsForCoordinator}
-          />
-        </Route>
+        {canViewOpenAppointments && (
```
yaron1m, line 125: Do you think this is needed?
Do we need to block users from loading these pages?

I mean, if they manually go to that url without having permissions to do anything in it, I am ok with it.
I just think this makes this code a bit too complex

harelglik: I think it's good practice to not allow unauthorized people to manually navigate to a page they shouldn't see.
It maybe complicates the code (a tiny bit) but makes us not need to think about it.

PR #189, opened 2021-06-05
Shorter Questionnaire + Error when appointment is already booked

yaron1m:


yaron1m:
Shorter questions:
image


Error message:
image

packages/donor/src/screens/questionnaire/QuestionnaireScreen.tsx
```diff
@@ -236,50 +126,20 @@ export default function QuestionnaireScreen({
           if (hasAlreadyDonated === "no") {
             setHasAlreadyDonated("");
           }
-          if (isWeightOver55 === "no") {
-            setIsWeightOver55("");
+          if (isWeightValid === "no") {
```
evbambly, line 129: Why are we comparing to a string and not bool or enum here?

packages/donor/src/screens/questionnaire/QuestionnaireScreen.tsx
```diff
@@ -155,28 +76,17 @@ export default function QuestionnaireScreen({
...
       value={isRightAge}
       onChange={setIsRightAge}
-      label={"האם אתה בטווח הגילאים 17-65?"}
+      label={"האם הנך מעל גיל 17?"}
     />
   );
 
-  const [isPregnantEver, setIsPregnantEver] = useState("");
-  const IsPregnantEver = (
+  const [wasPregnant, setWasPregnantEver] = useState("");
```
evbambly, line 83: Why is this a string value?

evbambly (approved):

packages/donor/src/screens/questionnaire/QuestionnaireScreen.tsx
yaron1m: I am not sure, but I didn't change this.
We can change this in a different PR

packages/donor/src/screens/questionnaire/QuestionnaireScreen.tsx
yaron1m: I am not sure, but I didn't change this.
We can change this in a different PR

PR #195, opened 2021-06-11
Input v2

yaron1m:


yaron1m:
image

image

packages/donor/src/components/basic/Input/Input.tsx
```diff
@@ -9,70 +8,48 @@ export enum InputVariant {
...
-  mainIcon,
-  actionIcon,
-  isDisabled,
   errorMessage,
   variant = InputVariant.standard,
-  isFullWidth = false,
-  onKeyPress = (ev) => undefined,
+  onSubmit,
 }: InputProps) {
-  let adornments = {
```
evbambly, line 46: Why are we removing the adronments?

evbambly (approved):

packages/donor/src/components/basic/Input/Input.tsx
yaron1m: No one uses it in V1. If we ever need it we can add it. Don't want to have הכנה למזגן in basic components

PR #200, opened 2021-06-12
Storybook/table

evbambly:
Storybook for Coordinator Table


packages/coordinator/src/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,263 @@
...
+      bloodType: BloodType.A_MINUS,
+      lastName: "333",
+      phone: "21112234",
+    },
+  },
+  {
+    rowData: {
+      bloodType: BloodType.A_PLUS,
+      phone: "0525938307",
+      email: "harelglik@gmail.com",
```
yaron1m, line 128: Please obfuscate all the data here.
i.e. change names, emails and phone numbers to not have any real info (like this one above)

yaron1m (approved):

packages/coordinator/src/components/Table/Table.stories.tsx
yaron1m: Do this for all the hard coded data

PR #204, opened 2021-06-17
Wizard screen

talwag:


packages/donor/src/components/Wizard/WizardScreen.module.scss
```diff
@@ -0,0 +1,26 @@
+:root {
```
yaron1m, line 1: What does `:root` do? Does this affect the entire app?

packages/donor/src/components/basic/Stepper/Stepper.tsx
```diff
@@ -1,52 +1,22 @@
...
-        <Button
-          size="small"
-          onClick={props.handleBack}
-          disabled={props.step === 0}
-        >
-          <KeyboardArrowRight />
-        </Button>
-      }
+      nextButton={null}
+      backButton={null}
```
yaron1m, line 19: Why null?

Don't we want to allow users to go back / forward using the arrows?

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
-        step={props.currentStep}
-        numberOfSteps={props.totalNumberOfSteps}
-        handleNext={props.onNext}
-        handleBack={props.onPrev}
-      />
-      <Button
-        onClick={props.onClick}
-        title={props.buttonText}
-        variant={props.buttonVariant}
+      <img src={props.pages[currentStep].imageUrl} alt="" />
```
yaron1m, line 55: You may want to limit the image size here

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
+import MobileStepper from "@material-ui/core/MobileStepper";
+import { BrowserRouter, Link } from "react-router-dom";
+import { PropTypes } from "@material-ui/core";
+import { ButtonVariant } from "../basic/Button";
 
-export interface WizardScreenProps {
-  imageUrl: string;
-  title: string;
-  content: string;
+const useStyles = makeStyles({
```
yaron1m, line 11: I don't think we should use makeStyles here, as we don't use it anywhere in the app.

Please see how we use styles in other components (by importing the css file to a variable) and do the same

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
+  root: {
+    flexDirection: "column",
+    fontFamily: "Rubik",
```
yaron1m, line 14: The font should be set in the app, and not here. 
If we ever want to change the app font, we should only change it in one place

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
+    margin: "8px",
+  },
 
-  // Action button
+  dots: {
+    margin: "30px 0",
+  },
+
+  dotActive: {
+    backgroundColor: "#A2D45E",
```
yaron1m, line 27: What is this color? Should we use the green we have in other places?

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
+  },
+});
+
+interface page {
+  imageUrl: string;
+  title: string;
+  content: { text: string; bold: boolean }[];
```
yaron1m, line 34: Maybe bold should be optional (and thus false by default)

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
   buttonText: string;
   buttonVariant: ButtonVariant;
-
-  onNext?: () => void;
-  onPrev?: () => void;
-  onClick: () => void;
+  buttonColor: PropTypes.Color;
+}
+export interface WizardScreenProps {
+  pages: Array<page>;
```
yaron1m, line 40: maybe `pages: page[]`?

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
+  dots: {
+    margin: "30px 0",
+  },
+
+  dotActive: {
+    backgroundColor: "#A2D45E",
+  },
+});
+
+interface page {
```
yaron1m, line 31: Please rename to WizardPage

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
   buttonVariant: ButtonVariant;
-
-  onNext?: () => void;
-  onPrev?: () => void;
-  onClick: () => void;
+  buttonColor: PropTypes.Color;
+}
+export interface WizardScreenProps {
+  pages: Array<page>;
+  currentStep: number;
```
yaron1m, line 41: Why is this an external prop?

Should we handle the count inside the component?

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
+  linkRow?: { text: string; linkText: string; link: string };
```
yaron1m, line 42: I am not sure what `linkRow` is. Can you explain?

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,40 +1,103 @@
...
+  callBack: () => void;
```
yaron1m, line 43: Maybe rename to `onFinish`?

packages/donor/src/components/Wizard/WizardScreen.module.scss
talwag: changed it to  .container

packages/donor/src/components/basic/Stepper/Stepper.tsx
talwag: for next button i made a different one using our button component
the design doesn't have back button

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: done

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: i couldn't make it work through the css file
i guess it had something to do with that those are inner mui mobilestepper elements

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: removed it

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: done

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: done

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: yeah i think so too, that was there before i thought maybe there a reason for it.

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: i used it to create the row with link to connect, but i changed it now to be part of the content array, i think it would be better this way.

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: done

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -28,55 +27,56 @@ const useStyles = makeStyles({
   },
 });
 
-interface page {
+interface wizardPage {
```
yaron1m, line 30: Types are CamelCased - `WizardPage`

yaron1m (approved):

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: thats the color in the design..

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: done

packages/donor/src/components/Wizard/WizardScreen.tsx
talwag: done

PR #205, opened 2021-06-22
Minor wizard changes

yaron1m:
image

### DIFF
--- a/packages/donor/src/components/Wizard/WizardScreen.module.scss
+++ b/packages/donor/src/components/Wizard/WizardScreen.module.scss
@@ -1,11 +1,10 @@
 .container {
-  height: 100vh;
+  height: 100%;
   display: flex;
   flex-direction: column;
   justify-content: center;
   align-items: center;
   text-align: center;
-  padding: 20px;
 }
 
 .image {
--- a/packages/donor/src/components/Wizard/WizardScreen.module.scss
+++ b/packages/donor/src/components/Wizard/WizardScreen.module.scss
@@ -1,22 +1,27 @@
 .container {
+  height: 100vh;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
   text-align: center;
-  color: #515151;
+  padding: 20px;
 }
 
 .image {
-  max-height: 200px;
+  max-height: 30vh;
 }
 
 .title {
-  color: #4caf50;
+  color: var(--donor-primary);
   font-size: 32px;
   font-weight: bold;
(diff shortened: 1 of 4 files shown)


packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,104 +1,88 @@
...
-                    {key.linkText}
-                  </Link>
-                </BrowserRouter>
-              ) : null}
-            </Text>
+
+              {key.button && (
+                <Button
+                  title={key.button.text}
+                  onClick={key.button.onClick!}
```
talwag, line 55: did you add ! on purpose?

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,104 +1,88 @@
...
+                  variant={ButtonVariant.text}
```
talwag, line 56: this button font is smaller then the rest of the content, it should be the same size

packages/donor/src/components/Wizard/WizardScreen.tsx
```diff
@@ -1,104 +1,88 @@
...
+    <ZMScreen className={styles.container}>
       <img
         className={styles.image}
         src={props.pages[currentStep].imageUrl}
         alt=""
       />
-      <Text className={styles.title}>{props.pages[currentStep].title}</Text>
+      <div className={styles.title}>{props.pages[currentStep].title}</div>
       <div className={styles.content}>
         {props.pages[currentStep].content.map((key, index) => (
```
talwag, line 41: i have missed that before, there is a console error since not all elements inside the loop has a special key, did it only to one of the elements

packages/donor/src/components/Wizard/WizardScreen.module.scss
```diff
@@ -1,30 +1,58 @@
 .container {
+  height: 100%;
```
talwag, line 2: for some reason now on devtools in the smaller devices the next button is out of the screen and i can't scroll down to it

packages/donor/src/components/Wizard/WizardScreen.tsx
yaron1m: Right, removed

packages/donor/src/components/Wizard/WizardScreen.module.scss
yaron1m: I will fix it in a different commit

PR #215, opened 2021-07-06
changed donation info date formatting

harelglik:
image


yaron1m:
Look at the order of the fields here:
image

yaron1m (approved):

PR #218, opened 2021-07-09
align date format in upcoming donation

harelglik:
image


packages/donor/src/utils/Date.consts.ts
```diff
@@ -0,0 +1,7 @@
+export const ShortDateFormat: Intl.DateTimeFormatOptions = {
```
yaron1m, line 1: Put this in Common.DateUtils

yaron1m (approved):

PR #222, opened 2021-07-10
Not a bug

evbambly:
image

It's just a feature of the table card ;)
https://trello.com/c/tEFRsE8L/159-ui-coordinator-planned-appointments-remove-separating-line-when-there-is-only-1-appointment

### DIFF
--- a/packages/coordinator/src/components/Table/CardTableItem.tsx
+++ b/packages/coordinator/src/components/Table/CardTableItem.tsx
@@ -50,7 +50,7 @@ export default function CardTableItem<T>({
     createStyles({
       root: {
         // From styles/layout/grid -> $gutter-vertical-small
-        marginBottom: ".5rem",
+        margin: ".5rem 0",
         // Overriding expanded:last-child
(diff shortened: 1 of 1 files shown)


packages/coordinator/src/components/Table/CardTableItem.tsx
```diff
@@ -50,7 +50,7 @@ export default function CardTableItem<T>({
     createStyles({
       root: {
         // From styles/layout/grid -> $gutter-vertical-small
-        marginBottom: ".5rem",
+        margin: ".5rem 0",
```
yaron1m, line 53: Nice, can you explain why this fixes the issue?

yaron1m (approved):

packages/coordinator/src/components/Table/CardTableItem.tsx
evbambly: It adds a margin on the top, 
bringing the card lower and showing it as a card with a lower and upper shadow-
instead of a divider

PR #223, opened 2021-07-17
Coordinator/show name

evbambly:
image


packages/functions/.firebaserc
```diff
@@ -1,5 +1,35 @@
 {
   "projects": {
-    "default": "blood-components"
+    "staging": "blood-components",
+    "default": "blood-components",
+    "production": "blood-components-9ad48"
+  },
+  "targets": {
```
yaron1m, line 7: Please remove "targets" and leave only "projects"

packages/functions/src/coordinator/GetCoordinatorHandler.ts
```diff
@@ -11,11 +12,15 @@ export default async function (
     console.error("Could not find calling user", callerId);
     throw Error(`User is not a coordinator`);
   }
+  const coordinatorDonorUser = await DonorDAL.getDonor(callerId);
```
yaron1m, line 15: Please run in parallel.

This means to call `DonorDAL.getDonor(callerId);` before you call `getCoordinator` but without the `await`, and only `await` on it later

yaron1m, line 15: Please run in parallel.

This means to call `DonorDAL.getDonor(callerId);` before you call `getCoordinator` but without the `await`, and only `await` on it later

packages/functions/src/testUtils/TestSamples.ts
```diff
@@ -12,10 +12,11 @@ export const sampleUser = {
   testUser: false,
 };
 
-export function saveTestDonor(donorId: string) {
+export function saveTestDonor(donorId: string, donorFields?: Partial<DbDonor>) {
```
yaron1m, line 15: This already saves first and last names, why not use them?

packages/functions/src/coordinator/GetCoordinatorHandler.ts
```diff
@@ -11,11 +12,15 @@ export default async function (
...
 
   return {
     coordinator: {
       role: coordinator.role,
       activeHospitalsForCoordinator: getCoordinatorHospitals(coordinator),
+      name: coordinatorDonorUser
```
yaron1m, line 21: Can this really be undefined?
Do we support coordinators who's donors have no names?

packages/coordinator/src/components/Header/CoordinatorHeader.tsx
```diff
@@ -20,12 +21,14 @@ interface CoordinatorHeaderProps {
...
   flags: HeaderButtonFlags;
   getEmail: () => string | undefined;
+  coordinator: Coordinator | undefined;
 }
 
 export default function CoordinatorHeader({
   flags,
   onSignOut,
   getEmail,
+  coordinator,
```
yaron1m, line 31: Can you add a story to CoordinatorHeader?

packages/functions/src/testUtils/TestSamples.ts
evbambly: I want to expect() a test-specific variable

packages/functions/src/coordinator/GetCoordinatorHandler.ts
evbambly: Safety?

yaron1m: I don't think this is needed, your call

packages/functions/src/testUtils/TestSamples.ts
yaron1m: Umm... ok.

You can just expect the existing first and last name, but I don't think this is critical

packages/functions/src/coordinator/GetCoordinatorHandler.ts
yaron1m: I see you marked as resolved, but I don't see any code changes

yaron1m (approved):

PR #231, opened 2021-08-03
Add sign in button to header / Task 300

idanlevi1:
https://trello.com/c/p3l8bXyY/300-add-sign-in-button-to-header 
image


packages/donor/src/components/AppHeader/style.module.scss
```diff
@@ -39,7 +39,17 @@
...
   margin-left: 15px;
-  margin-top: 5px;
+  display: flex;
+}
+
+.login_text {
+  color: var(--donor-link);
+  text-align: center;
+  font-size: 14px;
+  font-family: "Rubik";
```
yaron1m, line 49: Do you need to set it?
It should inherit Rubik by default, no?

yaron1m (approved):

packages/donor/src/components/AppHeader/style.module.scss
idanlevi1: You're right, I'm updating this

PR #238, opened 2021-08-13
Donor update whatsapp link for questionnaire / Task 337

idanlevi1:


packages/donor/src/screens/questionnaire/questions/QuestionsIchilov.tsx
```diff
@@ -102,7 +102,9 @@ export default function QuestionsIchilov({
         }
         buttonApproveText="שלח/י ואטסאפ לרכז שלך"
         onApproved={() => {
-          window.open(WHATSAPP_LINK);
+          window.open(
+            `${WHATSAPP_LINK}?text=אהלן, לפי השאלון באפליקציה אין לי יכולת לתרום טרומבוציטים כרגע, כיצד אוכל לשנות זאת?`
```
yaron1m, line 106: I suggest we create a util function that accepts a message and creates the WA link (instead of constructing the string in many places).

What do you think?

If you agree, I would put it in common cause it may be useful in the functions in the coming future

idanlevi1: I agree, take a look at the new changes.

yaron1m (approved):

yaron1m:
Great work!

PR #239, opened 2021-08-17
show wizard only once

shwinky:


packages/donor/src/navigation/AppRouter.tsx
```diff
@@ -151,7 +151,11 @@ export default function AppRouter() {
         />
         <Route
           path={"/" + MainNavigationKeys.OnboardingWizard}
-          render={() => <OnboardingWizardScreenContainer />}
+          render={() => {
+            if (localStorage.getItem("sawWizardScreen")) {
```
yaron1m, line 155: Please extract the string to a const

packages/donor/src/screens/onboarding/OnboardingWizardScreenContainer.tsx
```diff
@@ -4,6 +4,7 @@ import { MainNavigationKeys } from "../../navigation/app/MainNavigationKeys";
 
 export default function OnboardingWizardScreenContainer() {
   const history = useHistory();
+  localStorage.setItem("sawWizardScreen", "true");
```
yaron1m, line 7: Should you use `useEffect` to make sure this only runs once?

Also, this should run when the user *finishes* the wizard

yaron1m (approved):

yaron1m:
Very nice!

PR #244, opened 2021-08-30
added appointment status

shwinky:


packages/functions/tsconfig.json
```diff
@@ -1,3 +1,24 @@
 {
-  "extends": "../../tsconfig.json"
+  "extends": "../../tsconfig.json",
```
yaron1m, line 2: Revert this change please

packages/functions/src/coordinator/AddNewAppointmentsHandler.ts
```diff
@@ -44,6 +45,7 @@ function addAppointmentsToBatch(
     ),
     hospital: slotsRequest.hospital,
     donorId: "",
+    status: AppointmentStatus.AVAILABLE,
```
yaron1m, line 48: Please add this check to tests

packages/functions/src/donor/CancelAppointmentHandler.ts
```diff
@@ -36,6 +36,7 @@ export default async function (
 
   appointment.donorId = "";
   const updatedAppointment = removeDonorFromDbAppointment(appointment);
+  updatedAppointment.status = AppointmentStatus.AVAILABLE;
```
yaron1m, line 39: This change should be inside `removeDonorFromDbAppointment`.

This way it will also be used by `DeleteAppointmentsHandler` which is missing in this PR.

Also, please update the tests for `DeleteAppointmentsHandler` accordingly

yaron1m (approved):

PR #247, opened 2021-09-08
added appointment migration script

shwinky:
Not sure we should keep it in the code base...but after you approve we can cancel PR
I tested it with the unit-tests....maybe we should run it against test and see the results first....
also...wasn't really sure regarding the statuses....

### DIFF
--- a/packages/functions/src/dal/AppointmentDataAccessLayer.ts
+++ b/packages/functions/src/dal/AppointmentDataAccessLayer.ts
@@ -131,6 +131,22 @@ export async function getAvailableAppointments() {
   }));
 }
 
+export async function getAllLatestAppointments() {
+  const setDate = new Date();
+  setDate.setDate(setDate.getDate()-30);  // set date 30 days back
+
+  const appointments = (await admin
+    .firestore()
+    .collection(Collections.APPOINTMENTS)
+    .where("donationStartTime", ">", setDate)
+    .get()) as FirebaseFirestore.QuerySnapshot<DbAppointment>;
+
+  return appointments.docs.map((doc) => ({
+    id: doc.id,
+    ...doc.data(),
+  }));
+}
+
 export async function deleteAppointmentsByIds(appointmentIds: string[]) {
   const batch = admin.firestore().batch();
   const collection = admin.firestore().collection(Collections.APPOINTMENTS);
--- a/packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
+++ b/packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
@@ -1,0 +1,39 @@
+import { getAllLatestAppointments, setAppointment } from "../AppointmentDataAccessLayer";
+import { AppointmentStatus } from "@zm-blood-components/common/src";
+const fs = require('fs');
+
+export async function addStatusForAppointments(){
+  const appointments = await getAllLatestAppointments();
+
+  return appointments.map((appointment) => {
+    let newStatus = AppointmentStatus.AVAILABLE;
+    const now = new Date();
+    if(appointment.donorId !== "")
+    {
+      if (appointment.donationStartTime.toDate() < now)
+        newStatus = AppointmentStatus.COMPLETED; // should this be completed?
+      else
+        newStatus = AppointmentStatus.BOOKED;
+    }
+    else
+    {
+      if(appointment.donationStartTime.toDate() < now)
+        newStatus = AppointmentStatus.NOSHOW
+    }
+    return {
+      status: newStatus,
+      ...appointment
+    };
+  });
+}
+
+export async function migrateAppointmentStatus(){
+  const updatedAppointments = await addStatusForAppointments();
+  const stream = fs.createWriteStream('results.txt', {flags:'a'});
+  for(const appointment of updatedAppointments){
+    stream.write(`${appointment.id},${appointment.donorId},${appointment.donationStartTime.toDate()},${appointment.status}\n`);
+    await setAppointment(appointment)
+  }
+  stream.end()
+
+}


packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
```diff
@@ -0,0 +1,39 @@
+import { getAllLatestAppointments, setAppointment } from "../AppointmentDataAccessLayer";
+import { AppointmentStatus } from "@zm-blood-components/common/src";
+const fs = require('fs');
+
+export async function addStatusForAppointments(){
+  const appointments = await getAllLatestAppointments();
```
yaron1m, line 6: Why latest? Why not all?

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
```diff
@@ -0,0 +1,39 @@
...
+    {
+      if (appointment.donationStartTime.toDate() < now)
+        newStatus = AppointmentStatus.COMPLETED; // should this be completed?
+      else
+        newStatus = AppointmentStatus.BOOKED;
+    }
+    else
+    {
+      if(appointment.donationStartTime.toDate() < now)
+        newStatus = AppointmentStatus.NOSHOW
```
yaron1m, line 21: This should be AVAILABLE, no?
I think noShow is when it is booked but the donor did not show up

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+    return {
+      status: newStatus,
+      ...appointment
```
yaron1m, line 25: You should switch these 2 lines.

Otherwise if the appointment already has a status it won't be updated

Or is it inetntional?

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
```diff
@@ -0,0 +1,39 @@
...
+    };
+  });
+}
+
+export async function migrateAppointmentStatus(){
+  const updatedAppointments = await addStatusForAppointments();
+  const stream = fs.createWriteStream('results.txt', {flags:'a'});
```
yaron1m, line 32: Where is this file saved? 
Why is this needed? 
Can logs be enough?

yaron1m (commented):
Please note @evbambly is working on a framework to run jobs like this (see #248 )

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
shwinky: just thought that maybe all would be a big data set....but if you think we can run it on the whole database then we can...

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
shwinky: can an appointment be AVAILABLE if the appointment date has passed?

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
shwinky: it was intentional.... it was a fallback that if there was already a status it will take the status...

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
shwinky: didn't see we used log anywhere... I guess it would work...I just wanted a single file to give me the result...it would be saved in the folder the script would run from....

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
yaron1m: We have <500 entries in the DB, running it on all should be fine

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
yaron1m: Yes, this is how we identify appointments that were not booked

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
yaron1m: Cool

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
yaron1m: See log line example here:
https://github.com/startach/ZM-BloodComponents/blob/2f5fa1ebe6a2e83d1ab434c8cf8eaed4db0e05c4/packages/functions/src/RequestHandleWrapper.ts#L18

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
```diff
@@ -1,39 +1,39 @@
...
-        newStatus = AppointmentStatus.COMPLETED; // should this be completed?
-      else
-        newStatus = AppointmentStatus.BOOKED;
-    }
-    else
-    {
-      if(appointment.donationStartTime.toDate() < now)
-        newStatus = AppointmentStatus.NOSHOW
+        newStatus = AppointmentStatus.COMPLETED;
+      // should this be completed?
```
yaron1m, line 18: That depends on when would we run this job.
Currently there's no way to mark an appointment as COMPLETED, so I think we should mark all appointments with a donor id with one status, probably `Booked`.

Once we roll out to production a way to confirm a donation we can go back to all past booked appointments and change them to COMPLETED.

What do you think?

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
```diff
@@ -1,39 +1,39 @@
...
+  setAppointment,
+} from "../AppointmentDataAccessLayer";
 import { AppointmentStatus } from "@zm-blood-components/common/src";
-const fs = require('fs');
+import * as functions from "firebase-functions";
+const fs = require("fs");
 
-export async function addStatusForAppointments(){
-  const appointments = await getAllLatestAppointments();
+export async function addStatusForAppointments() {
```
yaron1m, line 9: What triggers this function?

Eitan has finished his job framework, you can integrate into that

packages/functions/src/dal/migrationScripts/ChangeAppointmentStatusMigration.ts
shwinky: I think its a little duplicate work....how does it hurt us that these are COPMLETED now? this would just add another migration, although it has no meaning now...
BUT, I'll change it so we can roll

packages/functions/src/dal/AppointmentDataAccessLayer.ts
```diff
@@ -131,6 +131,20 @@ export async function getAvailableAppointments() {
   }));
 }
 
+export async function getAllAppointments() {
+  const setDate = new Date();
+  const appointments = (await admin
+    .firestore()
+    .collection(Collections.APPOINTMENTS)
+    .where("donationStartTime", ">", setDate)
```
yaron1m, line 139: This will return only future appointments, no?

yaron1m: I would change the > to < and call this function `getPastAppointments`

yaron1m:
Great work!
Let's consult with Nir on how and when to run this on STG

yaron1m (approved):

PR #248, opened 2021-09-09
Function/job handler

evbambly:
Implementing a generic JobHandler to handle DB migrations and other operations needs.


packages/functions/src/jobs/SchemaJobs.ts
```diff
@@ -0,0 +1,39 @@
...
+    message: "",
+    status: 400,
+  };
+
+  switch (request.query?.jobName) {
+    case Jobs.DbMigrationAddAppointmentStatus:
+      response = await DbMigrationAddAppointmentStatus();
+      break;
+    default:
+      res.send(404);
```
yaron1m, line 28: You need to `return` after this, otherwise you would send `400` and not `404`

yaron1m (approved):

PR #250, opened 2021-09-14
Initial Commit

evbambly:
Added CompletedDonationsCount to DbDonor && Donor


packages/common/src/types.ts
```diff
@@ -51,6 +52,7 @@ export type Donor = {
   phone: string;
   bloodType: BloodType;
   notificationSettings: DonorNotificationSettings;
+  completedDonationsCount: number;
```
yaron1m, line 55: We said this is not needed yet, no?

yaron1m (commented):
Can you please update the commit message?

Also, looks like this is not ready for merge yet

PR #253, opened 2021-09-24
Add copy appointments details to clipboard button

yedidya03:
changes in this PR : 
  - fixing copied data
  - fixing table alignment

things that need to fixred :
  - streching the line of the table labels
  - fixing alignment in mobile size (not better in previous versions)
Capture

### DIFF
--- a/packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
+++ b/packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
@@ -31,13 +31,13 @@ const GetExpandedColumns = (
     cellRenderer: ({ donorName, booked }) => {
       return booked && donorName ? donorName : "אין רישום";
     },
-    hideIfNoData: true,
+    hideIfNoData: false,
   },
   {
     label: "סוג דם",
     cellRenderer: ({ bloodType }) =>
       bloodType ? LocaleUtils.getBloodTypeTranslation(bloodType) : null,
-    hideIfNoData: true,
+    hideIfNoData: false,
   },
   {
     label: "טלפון",
@@ -49,7 +49,7 @@ const GetExpandedColumns = (
           </a>
         </div>
       ),
-    hideIfNoData: true,
+    hideIfNoData: false,
   },
   {
     label: "נקבע בתאריך",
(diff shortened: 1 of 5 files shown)


packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
```diff
@@ -190,14 +196,20 @@ export const MainAppointmentTableColumns = (
...
-      return (
+      var visable =
         !showOnlyRecentChanges &&
         appointments.find(
           (a) => a.recentChangeType || a.recentChangeType === 0
-        ) && <Chip colorScheme={ChipColorScheme.New} label="חדש" />
+        );
+
+      return (
+        <div style={{ visibility: visable ? "visible" : "hidden" }}>
```
yaron1m, line 208: I don't like the inline styling.

You could return null if visible is false

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
```diff
@@ -190,14 +196,20 @@ export const MainAppointmentTableColumns = (
       appointments.filter((a) => a.booked).length,
   },
   {
+    label: "",
     cellRenderer: ({ appointments }) => {
-      return (
+      var visable =
```
yaron1m, line 201: const

yaron1m (approved):

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
yedidya03: I did that because in that way it still takes the place and the aligning remains the same..

yaron1m: You can return an empty div

yedidya03: but the width of the empty div wont be the same and it will cause missalignment.

yaron1m: ok

yaron1m (approved):

PR #254, opened 2021-09-25
use appointment status field

shwinky:
need to create an index in firebase for the new field so we can query by it...

### DIFF
--- a/packages/functions/src/dal/AppointmentDataAccessLayer.ts
+++ b/packages/functions/src/dal/AppointmentDataAccessLayer.ts
@@ -120,7 +120,6 @@ export async function getAvailableAppointments() {
   const appointments = (await admin
     .firestore()
     .collection(Collections.APPOINTMENTS)
-    .where("donorId", "==", "")
     .where("status", "==", AppointmentStatus.AVAILABLE)
     .where("donationStartTime", ">", now)
     .orderBy("donationStartTime")
--- a/packages/functions/src/dal/AppointmentDataAccessLayer.ts
+++ b/packages/functions/src/dal/AppointmentDataAccessLayer.ts
@@ -121,6 +121,7 @@ export async function getAvailableAppointments() {
     .firestore()
     .collection(Collections.APPOINTMENTS)
     .where("donorId", "==", "")
+    .where("status", "==", AppointmentStatus.AVAILABLE)
     .where("donationStartTime", ">", now)
     .orderBy("donationStartTime")
     .get()) as FirebaseFirestore.QuerySnapshot<DbAppointment>;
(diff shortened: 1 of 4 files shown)


packages/functions/src/dal/AppointmentDataAccessLayer.ts
```diff
@@ -121,6 +121,7 @@ export async function getAvailableAppointments() {
     .firestore()
     .collection(Collections.APPOINTMENTS)
     .where("donorId", "==", "")
+    .where("status", "==", AppointmentStatus.AVAILABLE)
```
yaron1m, line 124: I think this should replace the `donorId` check

packages/functions/src/donor/BookAppointmentHandler.ts
```diff
@@ -2,16 +2,10 @@ import { getDonorOrThrow, setDonor } from "../dal/DonorDataAccessLayer";
...
 } from "../dal/AppointmentDataAccessLayer";
 import * as admin from "firebase-admin";
-import {
-  AppointmentStatus,
-  BookingChange,
-  DbDonor,
-  FunctionsApi,
-  Hospital,
-} from "@zm-blood-components/common";
+import { AppointmentStatus, BookingChange, DbDonor, FunctionsApi, Hospital } from "@zm-blood-components/common";
```
yaron1m, line 8: This means you ran install in `function` folder and not in the outer one.

run `yarn install` in the root folder, and then `style` again

packages/functions/tsconfig.json
```diff
@@ -1,3 +1,24 @@
 {
-  "extends": "../../tsconfig.json"
+  "extends": "../../tsconfig.json",
```
yaron1m, line 2: Please remove these changes

packages/functions/src/donor/CancelAppointmentHandler.ts
```diff
@@ -3,7 +3,7 @@ import {
   removeDonorFromDbAppointment,
   setAppointment,
 } from "../dal/AppointmentDataAccessLayer";
-import { FunctionsApi } from "@zm-blood-components/common";
+import { AppointmentStatus, FunctionsApi } from "@zm-blood-components/common";
```
yaron1m, line 6: Is this needed?

yaron1m (commented):
The PR looks good but we can't merge it until we run the migration on all existing appointments, right?

yaron1m:
Also, I see tests are failing because there's no index on `status` and `donationStartTime`. You need to add it in STG and Prod

shwinky:
@yaron1m if only you read my first comment :)
need you help creating that index....since it don't know how to :)

yaron1m (approved):

PR #260, opened 2021-09-27
Add copy popup message

yedidya03:
adding  a popup message that indicates that the data was copied.

Untitled

### DIFF
--- a/packages/coordinator/src/components/Popup/Popup.module.scss
+++ b/packages/coordinator/src/components/Popup/Popup.module.scss
@@ -23,3 +23,12 @@ hr {
   font-weight: bold !important;
   color: #fff !important;
 }
+
+.flashMessage {
+  background: #008CBA;
+  padding: 15px;
+  border: none;
+  border-radius: 15px;
+  font-size: 24px;
+  color: white;
+}
--- a/packages/coordinator/src/components/Popup/Popup.module.scss
+++ b/packages/coordinator/src/components/Popup/Popup.module.scss
@@ -24,8 +24,15 @@ hr {
   color: #fff !important;
 }
 
+.flashMessageContainer {
+  z-index: 100;
+  position: absolute;
+  bottom: 50px;
+  left: 5%;
+}
+
 .flashMessage {
-  background: #008CBA;
+  background: #008cba;
   padding: 15px;
   border: none;
   border-radius: 15px;
--- a/packages/coordinator/src/components/Popup/PopupFlashMessage.tsx
+++ b/packages/coordinator/src/components/Popup/PopupFlashMessage.tsx
@@ -1,24 +1,38 @@
-import { Fade } from '@material-ui/core';
+import { Fade } from "@material-ui/core";
 
 type PopupFlashMessageProps = {
-    message : string, 
-    showFlash  : boolean,
-    styleType? : string,
+  message: string;
+  showFlash: boolean;
+  styleType?: string;
 };
 
-export default function PopupFlashMessage ( {
-    message,
-    showFlash,
-    styleType,
+export default function PopupFlashMessage({
+  message,
+  showFlash,
+  styleType,
 }: PopupFlashMessageProps) {
-
-    return (
-        <div style={{zIndex: 100 ,position: 'absolute', bottom: '50px', left: '100px'}}>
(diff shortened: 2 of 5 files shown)


packages/coordinator/src/components/Popup/PopupFlashMessage.tsx
```diff
@@ -0,0 +1,20 @@
+import { Fade } from "@material-ui/core";
+import styles from "./Popup.module.scss";
+
+type PopupFlashMessageProps = {
+  message: string;
+  showFlash: boolean;
```
yaron1m, line 6: This can work, but I think it would be better to pass `showFlash` and `onHide` props to this component.

In my opinion the timeout should be here and not in the calling class.

Imagine we want to use this flash popup in another place, I don't want to write `setTimeout` twice

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsScreen.tsx
```diff
@@ -156,6 +171,8 @@ export default function ManageAppointmentsScreen({
         onApproved={popupData.onApproved}
         onClose={() => setPopupData(emptyPopupData)}
       />
+      <PopupFlashMessage message={flashMessage} showFlash={showFlash} />
```
yaron1m, line 174: I think the message can be hard coded here, as  there's only one option for it - "הפרטים הועתקו בהצלחה"

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
```diff
@@ -19,12 +19,14 @@ import Chip, { ChipColorScheme } from "../../components/Chip";
 import userMinusIcon from "../../assets/user-minus.svg";
 import trashIcon from "../../assets/trash.svg";
 import copyIcon from "../../assets/copy.svg";
+import { Fade } from "@material-ui/core";
```
yaron1m, line 22: Is this used?

packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
```diff
@@ -130,6 +135,11 @@ const GetExpandedColumns = (
 
       return (
         <div className={Styles["icons-container"]}>
+          <div>
+            <Fade in={false} timeout={{ enter: 300, exit: 1000 }}>
+              <h1>Header</h1>
+            </Fade>
+          </div>
```
yaron1m, line 142: Is this still needed?

packages/coordinator/src/components/Popup/PopupFlashMessage.tsx
yedidya03: I partially solved it. now the timeout algorithm is implemented in the component, but still there are two different parameters (message and showFlash). Having it all on the message parameter made a little mess because then the message was empty during the fade out. so in the meanwhile I prefered that it work.

packages/coordinator/src/components/Popup/Popup.module.scss
```diff
@@ -23,3 +23,19 @@ hr {
...
+
+.flashMessageContainer {
+  z-index: 100;
+  position: fixed;
+  bottom: 50px;
+  left: 5%;
+}
+
+.flashMessage {
+  background: #555555;
```
yaron1m, line 35: Extract color to a variable in `variables.scss`

packages/coordinator/src/components/Popup/PopupFlashMessage.tsx
```diff
@@ -0,0 +1,28 @@
...
+export default function PopupFlashMessage({
+  message,
+  showFlash,
+  hide,
+}: PopupFlashMessageProps) {
+  if (showFlash) {
+    setTimeout(() => {
+      hide();
+    }, 700);
+  }
```
yaron1m, line 19: This `if` should be in a `useEffect`, this way it won't be triggered too many times. It should depend on `showFlash` and `hide`

yaron1m (approved):

PR #261, opened 2021-09-28
CronJobs to send confirmation emails

evbambly:


packages/functions/src/dal/AppointmentDataAccessLayer.ts
```diff
@@ -131,6 +131,33 @@ export async function getAvailableAppointments() {
   }));
 }
 
+export async function getAppointmentsByStatus(
+  status: AppointmentStatus,
+  fromTime?: Date,
+  toTime?: Date
```
yaron1m, line 137: Please follow the pattern in line 58
`options: { fromTime?: Date; toTime?: Date }`

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const SendConfirmationReminders = async (from: Date, to: Date) => {
+  const appointments = await getAppointmentsByStatus(
+    AppointmentStatus.BOOKED,
+    from,
+    to
+  );
+
+  const donorsInAppointments = await getDonors(
+    appointments.map((appointment) => appointment.donorId)
```
yaron1m, line 50: Do you need to distinct the users somehow?

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -0,0 +1,69 @@
...
+  );
+
+  appointments.forEach((appointment) => {
+    const donor = donorsInAppointments.find(
+      (donor) => donor.id == appointment.donorId
+    );
+
+    if (donor == undefined) {
```
yaron1m, line 58: Either `if(!donor)` or `if(donor === undefined)`

packages/functions/src/notifications/NotificationSender.ts
```diff
@@ -52,8 +56,12 @@ export async function sendEmailToDonor(
     templateId: templateId,
     dynamicTemplateData: data,
   };
-
-  addCalendarEventToDonor(msg, type, data, donor.email);
+  if (
+    type == NotificationToDonor.APPOINTMENT_BOOKED ||
+    type == NotificationToDonor.APPOINTMENT_CANCELLED_BY_COORDINATOR
```
yaron1m, line 61: I think this should be a switch/case, in an external function.

i.e. `shouldAddCalendarEvent(type)`

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as functions from "firebase-functions";
+import { getAppointmentsByStatus } from "../dal/AppointmentDataAccessLayer";
+import { getDonors } from "../dal/DonorDataAccessLayer";
+import { getAppointmentNotificationData } from "../notifications/AppointmentNotificationData";
+import {
+  NotificationToDonor,
+  sendEmailToDonor,
+} from "../notifications/NotificationSender";
+
+export const ConfirmationReminderOnSameDay = functions.pubsub
```
yaron1m, line 11: I know it is not going to be easy, but please find a way to write tests for this job.

Let me know if you need help with it

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -0,0 +1,69 @@
...
+  .schedule("every 1 hours")
+  .onRun(async () => {
+    const topOfThisHourInMillis = new Date().setMinutes(0, 0, 0);
+    const oneHourInMillis = 1000 * 60 * 60;
+    // Offset by 1 ms, to avoid catching XX:00 twice
+    const topOfPreviousHourInMillis =
+      topOfThisHourInMillis - oneHourInMillis + 1;
```
yaron1m, line 18: So let's say the time now is 9:00.

Are you going to send pushes for appointments between 8:00:00.001 to 9:00?

In case this is correct I suggest adding a comment to make this clearer

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -0,0 +1,69 @@
...
+
+// Run every day at 11 AM Israel Time
+//https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+export const ConfirmationReminderOnNextDay = functions.pubsub
+  .schedule("* 11 * * *")
+  .timeZone("Asia/Jerusalem")
+  .onRun(() => {
+    const midnightOfToday = new Date().setHours(0, 0, 0, 0);
+    const oneDayInMillis = 1000 * 60 * 60 * 24;
+    const midnightOfYesterday = midnightOfToday - oneDayInMillis;
```
yaron1m, line 34: No need for 1ms buffer here as well?

packages/functions/src/dal/AppointmentDataAccessLayer.ts
evbambly: getAppointmentsByHospital also not consistent?

packages/functions/src/jobs/CronJobs.ts
evbambly: Why would it matter?

yaron1m: Efficiency?

packages/functions/src/dal/AppointmentDataAccessLayer.ts
```diff
@@ -131,6 +131,33 @@ export async function getAvailableAppointments() {
...
+  }
+  request = request.orderBy("donationStartTime");
+
+  const appointments =
+    (await request.get()) as FirebaseFirestore.QuerySnapshot<DbAppointment>;
+
+  return appointments.docs.map((doc) => ({
+    id: doc.id,
+    ...doc.data(),
+  }));
```
yaron1m, line 158: Please use the `toDbAppointments` I just pushed

packages/functions/src/dal/AppointmentDataAccessLayer.ts
yaron1m: Right. 

What ever you think is best

packages/functions/src/jobs/CronJobs.ts
evbambly: How would you suggest clarifying line #16?

packages/functions/src/jobs/CronJobs.ts
evbambly: Are we expecting donations at midnight?

packages/functions/src/jobs/CronJobs.ts
yaron1m: My comment wasn't about line 16 but was referring to these calculations.

I think something like this could be useful:

`A job triggered at 09:00 will send emails to all appointments with donationStartTime between  8:00:00.001 to 9:00`

packages/functions/src/jobs/CronJobs.ts
yaron1m: It could be, we should be prepared for it.

Also, I am not sure line 32 will result in midnight, because this depends on the timezone the job is running at.

I didn't comment on this before because I am ok with it running daily at 02:00 or something like it

packages/functions/src/jobs/CronJobs.ts
yaron1m: Did you change anything here or just marked it as resolved?

evbambly: Forgot to push

shwinky:
@yaron1m I added tests for the main function, let me know if you think its enough or do we need more tests
we are still missing the template for this email

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -0,0 +1,72 @@
...
+  const appointments = await getAppointmentsByStatus(
+    AppointmentStatus.BOOKED,
+    from,
+    to
+  );
+
+  const donorIds = appointments.map((appointment) => appointment.donorId);
+
+  const donorsInAppointments = await getDonors(
+    donorIds.filter((id, index) => donorIds.indexOf(id) === index)
```
yaron1m, line 53: Is this trying to remove duplications?

Consider using `_.uniq` - https://lodash.com/docs/4.17.15#uniq

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -0,0 +1,72 @@
...
+// Run every day at 11 AM Israel Time
+//https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+export const ConfirmationReminderOnNextDay = functions.pubsub
+  .schedule("* 11 * * *")
+  .timeZone("Asia/Jerusalem")
+  .onRun(() => {
+    const midnightOfToday = new Date().setHours(0, 0, 0, 0);
+    const oneDayInMillis = 1000 * 60 * 60 * 24;
+    //A job triggered at 00:00 will send emails to all appointments with donationStartTime between 00:00:00.001 (yesterday) to 00:00
+    const midnightOfYesterday = midnightOfToday - oneDayInMillis + 1;
```
yaron1m, line 35: This time calculating logic here (and above) is really error prone.

I suggest extracting it to a helper function that returns start and end time millis, and add simple unit tests to them to make sure we are getting the millis we expect

packages/functions/src/notifications/NotificationSender.ts
```diff
@@ -39,6 +40,9 @@ export async function sendEmailToDonor(
     case NotificationToDonor.APPOINTMENT_CANCELLED_BY_COORDINATOR:
       templateId = "d-efc1aeb750f542ec9fb1120c03e97a68";
       break;
+    case NotificationToDonor.DONATION_CONFIRMATION:
+      // TODO - Get template for donation confirmation
+      break;
```
yaron1m, line 45: Maybe `return` until implemented?

packages/functions/src/jobs/SendConfirmation.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+jest.mock("../notifications/NotificationSender");
+const mockedNotifier = mocked(sendEmailToDonor);
+
+const DONORS = ["BookedDonorId", "AvailableDonorId", "OutOfRangeDonorId"];
+const APPOINTMENTS = [
+  "BookedAppointment",
+  "AvailableAppointment",
+  "OutOfRangeAppointment",
+];
```
yaron1m, line 37: Add to the names something about the test name - this way we would know in the DB they belong to this test

packages/functions/src/jobs/SendConfirmation.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+afterEach(async () => {
+  await reset();
+});
+
+test("run on all appointments", async () => {
+  await createDonor(DONORS[0]);
+  await createDonor(DONORS[1]);
+  await createDonor(DONORS[2]);
+  const lastHour = new Date();
```
yaron1m, line 58: I believe test should not use `new Date()`, as this makes them dependent on the running time.

One option to solve this is make `SendConfirmationReminders` also accept `now` date, and set it here in the test to a constant value

packages/functions/src/jobs/SendConfirmation.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    admin.firestore.Timestamp.fromDate(lastHour)
+  );
+
+  await saveAppointment(
+    DONORS[1],
+    APPOINTMENTS[1],
+    AppointmentStatus.AVAILABLE,
+    admin.firestore.Timestamp.fromDate(lastHour)
+  );
+  const yesterday = new Date();
```
yaron1m, line 73: Continuing mu comment from above, here is a nice example - 
Theoretically if this test runs at 23:59:59 this could fail, as `lastHour` and `yesterday` would be on the same day as they are calculated 1s after the other

yaron1m (commented):
Sorry for taking (too) long to respond

shwinky:
@yaron1m I changed all the CronJob scheduling(it was also very complicated and buggy)...please look at to see if it makes sense

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -7,37 +7,30 @@ import {
...
 
-    SendConfirmationReminders(
-      new Date(topOfPreviousHourInMillis),
-      new Date(topOfThisHourInMillis)
-    );
+    await SendConfirmationReminders(lastHour, now);
   });
 
-// Run every day at 11 AM Israel Time
+// Run every day at 00:00 AM Israel Time
```
yaron1m, line 23: Why 00:00?

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -7,37 +7,30 @@ import {
...
+import _ from "lodash";
 
 export const ConfirmationReminderOnSameDay = functions.pubsub
-  .schedule("every 1 hours")
+  .schedule("0 * * * *")
+  .timeZone("Asia/Jerusalem")
   .onRun(async () => {
-    const topOfThisHourInMillis = new Date().setMinutes(0, 0, 0);
-    const oneHourInMillis = 1000 * 60 * 60;
-    //A job triggered at 09:00 will send emails to all appointments with donationStartTime between 8:00:00.001 to 9:00
```
yaron1m: This comment was important, we should still follow it

yaron1m:
Did you take a look at my comments from the last review (from 8 days ago)?

PR #263, opened 2021-09-28
Move util + add tests

yaron1m:


packages/functions/src/donor/CancelAppointmentHandler.ts
```diff
@@ -34,7 +34,8 @@ export default async function (
     console.error("Error notifying on cancelled appointment", appointment.id, e)
   );
 
-  const updatedAppointment = removeDonorFromDbAppointment(appointment);
+  const updatedAppointment =
+    DbAppointmentUtils.removeDonorFromDbAppointment(appointment);
```
evbambly, line 38: Doesn't this negate the concept of working with a DAL?

evbambly (approved):

packages/functions/src/donor/CancelAppointmentHandler.ts
yaron1m: I don't understand why this should be in the DAL.

It is not reading or writing any data, just mutating an object.

This is a classic util.

What do you think? Should this be in the DAL? Why?

evbambly: Mutating is CRUD, I understand DAL === CRUD?

If there should be any complex permission handling for specific mutating operations,
they should be handled directly and not in a util file, no?

yaron1m: > Mutating is CRUD, I understand DAL === CRUD?

🤷 

> If there should be any complex permission handling

What permissions handling?

PR #268, opened 2021-10-10
added route to complete appointments

shwinky:


packages/functions/src/notifications/CompleteAppointmentApi.ts
```diff
@@ -0,0 +1,25 @@
...
+  async (request, res) => {
+    if (!request.query.donorId) {
+      throw new Error("Invalid donor id");
+    }
+
+    if (!request.query.appointmentId) {
+      throw new Error("Invalid appointment id");
+    }
+    const donorId = request.query.donorId as string;
+    const appointmentId = request.query.appointmentId as string;
```
yaron1m, line 14: Extract the variables before the check

yaron1m (approved):

PR #270, opened 2021-10-15
Edit complete function

yedidya03:
changing complete appointment function to get a value on isNoshow

### DIFF
--- a/packages/functions/src/donor/CompleteAppointmentHandler.ts
+++ b/packages/functions/src/donor/CompleteAppointmentHandler.ts
@@ -4,30 +4,32 @@ import {
 } from "../dal/AppointmentDataAccessLayer";
 import { FunctionsApi } from "@zm-blood-components/common";
 import * as DbAppointmentUtils from "../utils/DbAppointmentUtils";
+import {
+  dbAppointmentToAppointmentApiEntry,
+  dbAppointmentToBookedAppointmentApiEntry,
+} from "../utils/ApiEntriesConversionUtils";
 
 export default async function (
   request: FunctionsApi.CompleteAppointmentRequest,
   callerId: string
-) {
+): Promise<FunctionsApi.CompleteAppointmentResponse> {
   const donorId = callerId;
 
   if (!request.appointmentId) {
     throw new Error("No appointment to complete");
   }
-  return {
-    appointment: await completeAppointmentFunc(
-      request.appointmentId,
-      donorId,
-      request.isNoshow
-    ),
-  };
+  return await completeAppointmentFunc(
+    request.appointmentId,
+    donorId,
+    request.isNoshow
+  );
 }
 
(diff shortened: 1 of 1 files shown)


yaron1m:
Is this ready for review?

yedidya03:
now it is

packages/functions/src/notifications/CompleteAppointmentApi.ts
```diff
@@ -13,8 +15,16 @@ export const completeAppointmentApi = functions.https.onRequest(
       throw new Error("Invalid appointment id");
     }
 
+    // if (!isNoshow || !(isNoshow in ["true", "false"]) ) {
```
shwinky, line 18: remove this

shwinky (approved):
approved, just fix the comments

packages/common/src/functions-api.ts
```diff
@@ -83,10 +83,7 @@ export interface CancelAppointmentRequest {
 export const CompleteAppointmentFunctionName = "completeAppointment";
 export interface CompleteAppointmentRequest {
   appointmentId: string;
-}
-
-export interface CompleteAppointmentResponse {
```
yaron1m, line 88: Why was the response deleted?

packages/functions/src/notifications/CompleteAppointmentApi.ts
```diff
@@ -13,8 +15,16 @@ export const completeAppointmentApi = functions.https.onRequest(
...
+    //   throw new Error("Invalid isNoshow value");
+    // }
+
     try {
-      await completeAppointmentFunc(appointmentId, donorId);
+      await completeAppointmentFunc(
+        appointmentId,
+        donorId,
+        isNoshow === "false"
```
yaron1m, line 26: Shouldn't this be `isNoShow === "true"`?

packages/functions/src/donor/CompleteAppointmentHandler.test.ts
```diff
@@ -103,6 +103,40 @@ test("Valid request complete appointment", async () => {
   expect(appointment[0].creatorUserId).toEqual("creatorUserId");
 });
 
+test("Valid request complete appointment with true parameter", async () => {
```
yaron1m, line 106: 2 Small comments:
1. The name of the test is not explanatory enough. What does the true or false mean? Consider something like "With No Show".
2. The two tests you added are identical, except for the status. For tests like this you can use `test.each`. See an example here:
https://github.com/startach/ZM-BloodComponents/blob/fa2be91d73aa46eac0e2a4db6976aafd464ef44b/packages/functions/src/coordinator/DeleteAppointmentsHandler.test.ts#L101-L134

yaron1m (commented):
Please make sure you also address my comments before merging

PR #276, opened 2021-10-20
Add copy appointments details to clipboard button

yedidya03:
adding the time to the copy.

### DIFF
--- a/packages/coordinator/src/screens/manageAppointmentsScreen/CopyAppointmentDetailsHelper.test.ts
+++ b/packages/coordinator/src/screens/manageAppointmentsScreen/CopyAppointmentDetailsHelper.test.ts
@@ -28,7 +28,7 @@ test.each([
   ["", undefined, ""],
   [undefined, "", ""],
 ])(
-  "test copy string provider without time value",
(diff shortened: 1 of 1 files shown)


packages/coordinator/src/screens/manageAppointmentsScreen/ManageAppointmentsTableConfig.tsx
```diff
@@ -95,10 +95,12 @@ const GetExpandedColumns = (
           tooltip: "העתק",
           iconUrl: copyIcon,
           onClick: () => {
-            let bookingDate = appointment.donationStartTimeMillis
-              ? DateUtils.ToDateString(appointment.donationStartTimeMillis)
+            const bookingDate = appointment.donationStartTimeMillis
+              ? DateUtils.ToDateString(appointment.donationStartTimeMillis) +
+                " " +
+                DateUtils.ToTimeString(appointment.donationStartTimeMillis)
```
yaron1m, line 101: Why is the start time here twice?

Also, please extract this `appointment -> copy string` to a function in an external file and add tests to it

PR #284, opened 2021-11-19
Add Hadasa Ein Kerem hospital

yaron1m:


packages/donor/src/screens/questionnaire/questions/QuestionsHadasaEinKerem.tsx
```diff
@@ -0,0 +1,120 @@
...
+  ]);
+
+  const isWrongAnswerChosen =
+    isWeightValid === false ||
+    isSurgeryValid ||
+    isTattooValid ||
+    isRightAge === false ||
+    wasPregnant === "yes";
+
+  return (
```
shwinky, line 49: I'm guessing these questions aren't special to Hadassah... wouldn't we want to create a common module with them ?

shwinky (approved):

packages/donor/src/screens/questionnaire/questions/QuestionsHadasaEinKerem.tsx
yaron1m: Every hospital has their special questions. Most are common but each hospital has their own changes.

How would you suggest doing that?

PR #286, opened 2021-11-22
changed startach logo

shwinky:


packages/donor/src/screens/authentication/AuthLoadingScreen.tsx
```diff
@@ -1,7 +1,7 @@
 import styles from "./AuthLoadingScreen.module.scss";
 import { Player } from "@lottiefiles/react-lottie-player";
 
-import startachLogo from "../../assets/mobile/startach_logo.svg";
+import startachLogo from "../../assets/mobile/startach_logo.png";
```
yaron1m, line 4: SVG is better than PNG as it is a vector file and has no resolution issues.

Can you export the new logo as SVG?

yaron1m (approved):

PR #287, opened 2021-11-22
adding manual appointment to getCoordinatorApoointments function

yedidya03:


packages/common/src/functions-api.ts
```diff
@@ -41,6 +41,7 @@ export type AppointmentApiEntry = {
 
   // If booked
   donorId?: string;
+  assigningCoordinator?: string;
```
yaron1m, line 44: Add a comment saying this only applies when donor is anonymous

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
```diff
@@ -96,10 +97,15 @@ async function filterDonorsInGroup(
...
+  assigningCoordinator: string
 ) {
   const donorIds = new Set(donors.map((donor) => donor.id));
   return appointments.filter(
-    (appointment) => appointment.donorId && donorIds.has(appointment.donorId)
+    (appointment) =>
+      appointment.donorId &&
+      (donorIds.has(appointment.donorId) ||
+        (appointment.donorId === "manual" &&
+          appointment.assigningCoordinator === assigningCoordinator))
```
yaron1m, line 109: Adding && and || in the same condition is bad practice.

Instead of using a one-liner, split this to if and else if...

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
```diff
@@ -96,10 +97,15 @@ async function filterDonorsInGroup(
...
+  donors: DbDonor[],
+  assigningCoordinator: string
 ) {
   const donorIds = new Set(donors.map((donor) => donor.id));
   return appointments.filter(
-    (appointment) => appointment.donorId && donorIds.has(appointment.donorId)
+    (appointment) =>
+      appointment.donorId &&
+      (donorIds.has(appointment.donorId) ||
+        (appointment.donorId === "manual" &&
```
yaron1m, line 108: Export this string to a const

yaron1m (approved):

PR #299, opened 2021-12-07
Change the donation process title according to userLogin state

ritabiniash:


packages/donor/src/screens/about/DonationProcessScreen.tsx
```diff
@@ -17,7 +19,9 @@ export default function DonationProcessScreen({
 
       <div className={styles.textContainer}>
         <div className={styles.title}>
-          תורמי ותורמות טרומבוציטים יקרים/ות שלום רב!
+          {isUserLoggedIn
+            ? "ירון, הנה מידע חשוב לקראת התרומה"
```
yaron1m, line 23: How would you handle a case where the donor name is not `ירון`?

ritabiniash: I can also pass the logged in user as a prop and render the name, or to manage a Login context that will handle the login store globally and consume it here rather than as a prop

yaron1m:
Closing this (as this was only for test)

PR #300, opened 2021-12-13
Merge branch 'add-coordinator-book-appointment-function' into add-cor…

yedidya03:
…dinator-book-non-manual


packages/functions/src/coordinator/CoordinatorBookAppointmentHandler.ts
```diff
@@ -7,13 +7,17 @@ export default async function (
   callerId: string
 ): Promise<FunctionsApi.BookAppointmentResponse> {
   if (request.donorId !== MANUAL_DONOR_ID) {
-    throw Error(`Coordinator booking for real donots is not supported yet`);
-  }
+    const donors = await CoordinatorDAL.getAdminDonors(callerId);
```
yaron1m, line 10: This is not very efficient. 
Why do you need to fetch ALL the coordinator donors, if you already know the donor id?

yedidya03: it is mainly for group coordinators that should be able to book only their donors..

yedidya03: however, what we can do is check the other coordinator types before doing that..

yaron1m: Let's say this is a group coordinator.

Why do you need to fetch all their donors? Why not just he one you want to validate?

yaron1m:
Any update here/

PR #303, opened 2021-12-17
adding option to get all appointments of coordinator in all hospitals

yedidya03:


packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
```diff
@@ -71,22 +71,33 @@ async function fetchCoordinator(callerId: string) {
   return coordinator;
 }
 
-async function validate(coordinator: DbCoordinator, hospital: Hospital) {
+async function validate(
```
yaron1m, line 74: Name is no longer right, as this not only validates.

How about `getValidHospitalsOrThrow`?

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
```diff
@@ -71,22 +71,33 @@ async function fetchCoordinator(callerId: string) {
...
+  coordinator: DbCoordinator,
+  hospital: Hospital | "all"
+): Promise<Hospital[]> {
   switch (coordinator.role) {
     case CoordinatorRole.SYSTEM_USER:
-      return;
     case CoordinatorRole.ZM_COORDINATOR:
     case CoordinatorRole.HOSPITAL_COORDINATOR:
+      if (hospital === "all") {
```
yaron1m, line 82: For SYSTEM_USER and ZM_COORDINATOR you should return all active hospitals

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
```diff
@@ -71,22 +71,33 @@ async function fetchCoordinator(callerId: string) {
...
         throw Error("Coordinator has no permissions for hospital");
       }
-      break;
+      return [hospital];
     case CoordinatorRole.GROUP_COORDINATOR:
       break;
   }
+
+  if (hospital === "all") {
+    return Object.values(Hospital);
```
yaron1m, line 97: Only ActiveHospitals should be returned

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
```diff
@@ -71,22 +71,33 @@ async function fetchCoordinator(callerId: string) {
...
         );
         throw Error("Coordinator has no permissions for hospital");
       }
-      break;
+      return [hospital];
     case CoordinatorRole.GROUP_COORDINATOR:
       break;
   }
+
+  if (hospital === "all") {
```
yaron1m, line 96: What flow reaches this code?

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yedidya03: it returns the hospital you ask for or all hospitals if ask for "all".. (see lines 96-100)

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yedidya03: done

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yedidya03: fixed

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yedidya03: fixed

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yaron1m: This pattern is prone to errors.

The switch-case should handle all roles, and throw for unfamiliar role. 
It should not return a valid response.

Imagine we decide to add a new role - Guest. That has no permissions at all.
We wouldn't want this role to automatically get permissions to all hospitals, right?

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yaron1m: I don't see the fix, where is it?

yedidya03: it passes the switch case and returns the ospitals that were asked or all the active hospitals

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yedidya03: got it, working on it..

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yaron1m: Are you sure?

Anyway, make sure you have tests that cover this

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.ts
yedidya03: done

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.test.ts
```diff
@@ -144,6 +148,59 @@ test("Valid request returns appointments of the right hospital", async () => {
   );
 });
 
+test("Valid request all returns appointments of the right hospital", async () => {
```
yaron1m, line 151: This test is great (although I would change the name to include testing HOSPITAL_COORDINATOR).

We also need a similar test for other roles to make sure they behave correctly when calling with "all".

For example, I would expect that SYSTEM_USER would also get the ICHILOV appointment for the same request.

Try to use `test.each()` to avoid code duplication

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.test.ts
```diff
@@ -144,6 +148,59 @@ test("Valid request returns appointments of the right hospital", async () => {
...
+  await saveAppointment(FUTURE_NOT_BOOKED, getDate(5), Hospital.TEL_HASHOMER);
+  const res = await callFunction(COORDINATOR_ID, undefined, true);
+  let appointments = res.appointments.filter((a) =>
+    ALL_APPOINTMENT_IDS.includes(a.id)
+  );
+  expect(appointments).toHaveLength(4);
+  expect(appointments[0].id).toEqual(PAST_BOOKED);
+  expect(appointments[1].id).toEqual(PAST_NOT_BOOKED);
+  expect(appointments[2].id).toEqual(FUTURE_BOOKED);
+  expect(appointments[3].id).toEqual(FUTURE_NOT_BOOKED);
```
yaron1m, line 194: You can use `expect(...).toContain(...)`

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.test.ts
```diff
@@ -244,16 +301,22 @@ async function createUser(role: CoordinatorRole, hospitals?: Hospital[]) {
...
     ? new Date().getTime() - earliestTimeDays * 24 * 60 * 60 * 1000
     : undefined;
 
-  const request: FunctionsApi.GetCoordinatorAppointmentsRequest = {
+  let request: FunctionsApi.GetCoordinatorAppointmentsRequest = {
     hospital: Hospital.TEL_HASHOMER,
     earliestStartTimeMillis,
   };
+
+  if (all) {
```
yaron1m, line 316: This is a bit confusing. Instead of having a default hospital and then overriding it from the arguments, I suggest you just ask for the hospital explicitly.

Even if this means you have to go over all existing calls and add TEL_HASHOMER to them

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.test.ts
yedidya03: done

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.test.ts
yedidya03: write

packages/functions/src/coordinator/GetCoordinatorAppointmentsHandler.test.ts
yedidya03: done

yaron1m (approved):
Nice work!

PR #304, opened 2021-12-19
Fix functions affected by manual donor

yedidya03:


packages/functions/src/coordinator/GetDonorsHandler.ts
```diff
@@ -6,6 +6,7 @@ export default async function (
   request: FunctionsApi.GetDonorsRequest,
   callerId: string
 ): Promise<FunctionsApi.GetDonorsResponse> {
+  // TODO : should it return also manual donor that this coordinator regisered?
```
yaron1m, line 9: I think not. They are not real entities

packages/functions/src/jobs/CronJobs.ts
```diff
@@ -52,12 +56,16 @@ export const SendConfirmationReminders = async (
...
   const donorIds = appointments.map((appointment) => appointment.donorId);
 
-  const donorsInAppointments = await getDonors(_.uniq(donorIds));
+  const donorsInAppointments: { [donorId: string]: DbDonor } = (
+    await getDonors(_.uniq(donorIds))
+  ).reduce((obj, donor: DbDonor) => {
+    return { ...obj, [donor.id]: donor };
+  }, {});
+
+  delete donorsInAppointments[MANUAL_DONOR_ID];
```
yaron1m, line 65: Use `donorIds.filter(id => id  !=== MANUAL_DONOR_ID)`

yaron1m, line 65: Use `donorIds.filter(id => id  !=== MANUAL_DONOR_ID)`

yaron1m, line 65: Use `donorIds.filter(id => id  !=== MANUAL_DONOR_ID)`

yedidya03: I added this code for improving time complexity in line 68..

yedidya03: however maybe it worth adding a comment of what those lines do..

packages/functions/src/coordinator/GetDonorsHandler.ts
yedidya03: 👍

packages/functions/src/jobs/CronJobs.ts
yaron1m: It is great, just don't use `delete`. 
Also, there's no need for `reduce`.

`.filter` will get you what you want, no?

yedidya03: In line 68, I want to get the DBDonor object from the donorId that I get from the appointment.
fot that, I need a dictionary like object that the key is the donorId and the value is the donor object, in that way I can get the donor object from the donorId in O(1) complexity.
so what I need to do is to build this dictionary..
what I have for building this dictionary is an array of DBDonor objects (from getDonor(...)) and inside eatch donor object there is a donorId that should turn to the key of the item in the dictionary.
filter will return an array, and that is not what I need. with reduce I can build this dictionary in an easy way.

yedidya03: about the delete, you are probebly write becouse the getDonors() function will not return items with id of MANUAL_DONOR_ID..

yaron1m: So why not use `map`?

```
const donorsInAppointments = await getDonors(_.uniq(donorIds));
const donorsMap : { [donorId: string]: DbDonor};
donorsInAppointments.map(donor => donorsMap[donor.id]=donor } 
```

yaron1m (approved):

PR #305, opened 2021-12-20
Add all hospitals in menu

yedidya03:


packages/common/src/HospitalUtils.ts
```diff
@@ -9,8 +9,18 @@ export const activeHospitals = [
...
-export function getAllHospitalOptions(defaultLabel?: string) {
-  return getHospitalOptions(activeHospitals, defaultLabel);
+export const ALL_HOSPITALS_SELECT = "ALL";
+
+export function getAllHospitalOptions(
+  hospitals: Hospital[],
+  allLable: string,
+  defaultLabel?: string
+): SelectOption<Hospital | typeof ALL_HOSPITALS_SELECT | "">[] {
+  let options: SelectOption<Hospital | typeof ALL_HOSPITALS_SELECT | "">[] =
```
yaron1m, line 19: I see we have `Hospital | typeof ALL_HOSPITALS_SELECT | ""` written in multiple places.

Maybe we should create a type for it:
`type HospitalOptionKey = Hospital | typeof ALL_HOSPITALS_SELECT | ""`

packages/common/src/HospitalUtils.ts
```diff
@@ -9,8 +9,18 @@ export const activeHospitals = [
   Hospital.TEL_HASHOMER,
 ];
 
-export function getAllHospitalOptions(defaultLabel?: string) {
-  return getHospitalOptions(activeHospitals, defaultLabel);
+export const ALL_HOSPITALS_SELECT = "ALL";
+
+export function getAllHospitalOptions(
```
yaron1m, line 14: 1. I don't think we need to take the label as an argument, but we need a flag to know whether to put the "all" and the "" options or not.
2. Not all users should see the option "All" , only those who have permissions to more than 1 hospital.

All in all, you can resolve this by just checking if `hospitals.length > 1` and if so, add the desired option.

I think you can _hard code_ the labels, no need to pass them.

packages/common/src/HospitalUtils.ts
yedidya03: done

packages/common/src/HospitalUtils.ts
yedidya03: it will break lots of code in different places that use this method. I think it is better to do this in a separate PR.

yedidya03: So what I did is for:
   1. I changed it to be a hard coded value, but only for the getALl function..
   2. I moved the check of more then one hospital into the getAll function (it was in the screen file)

packages/common/src/HospitalUtils.ts
```diff
@@ -9,8 +11,19 @@ export const activeHospitals = [
...
+  hospitals: Hospital[]
+): SelectOption<Hospital | typeof ALL_HOSPITALS_SELECT | "">[] {
```
yaron1m, line 18: Use `HospitalUtils.HospitalOptionKey`

yaron1m (approved):

packages/common/src/HospitalUtils.ts
yedidya03: write

PR #308, opened 2021-12-25
Common popup fix

ritabiniash:
Common coordinator pop up pop up 
image


packages/coordinator/src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,68 @@
...
+  DialogActions,
+  DialogContent,
+  makeStyles,
+} from "@material-ui/core";
+import Button, { ButtonVariant } from "../Button";
+
+export type PopupProps = {
+  open: boolean;
+  title?: string | React.ReactNode;
+  content?: string | React.ReactNode;
```
yaron1m, line 14: Do we currently need any title / content that is not string?

If not, I would keep it string for now, and change to ReactNode if we ever need it

packages/coordinator/src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onClose: (event: object, reason: string) => void;
```
yaron1m, line 15: What is the `reason` used for?

packages/coordinator/src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <Button
+            title={primaryButtonText}
+            onClick={onPrimaryButtonClick || noop}
+            color={"secondary"}
+          />
+        )}
+        {cancelButtonText && (
+          <Button
+            title={cancelButtonText}
+            onClick={onCancelButtonClick || noop}
```
yaron1m, line 60: Why is the `noop` needed?

packages/coordinator/src/components/Popup/Popup.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+  primaryButtonText: "הסר תורם",
+  cancelButtonText: "השאר תורם",
+};
+
+const Template: Story<PopupProps> = (args) => {
+  const [open, setOpen] = useState(false);
+  const close = async () => setOpen(false);
+
+  return (
+    <div style={{ margin: 20 }}>
```
yaron1m, line 22: Something I discovered recently - 
Instead of adding padding like this, you can add to the exported default in line 8:
`parameters: { layout: "padded" }`

packages/coordinator/src/components/Popup/Popup.tsx
ritabiniash: `onCancelButtonClick` is optional and without `noop`, the linter is mad.
So instead of creating the props for button in a separate code and check  `onCancelButtonClick`'s condition, we will pass `noop` , I think it is more readable.

packages/coordinator/src/components/Popup/Popup.tsx
ritabiniash: copied it from MUI `Dialog`'s signature, will remove

packages/coordinator/src/components/Popup/Popup.tsx
yaron1m: Got it, cool

yaron1m (approved):

PR #310, opened 2021-12-31
Add first name and last name validation in user profile

ritabiniash:
Task : https://trello.com/c/1gJuQDZV/380-sign-up-page-first-name-last-name-should-be-required-and-at-least-2-characters

Validate that first name and last name exist and have more than 2 characters in profile update page
image

### DIFF
--- a/packages/donor/src/screens/extendedSignup/PersonalDetails.tsx
+++ b/packages/donor/src/screens/extendedSignup/PersonalDetails.tsx
@@ -1,3 +1,4 @@
+import { isEmpty } from "lodash";
 import { BloodType, BloodTypeUtils } from "@zm-blood-components/common";
 import Button from "../../components/basic/Button";
 import Input from "../../components/basic/Input";
@@ -42,6 +43,7 @@ export default function PersonalDetails(props: PersonalDetailsProps) {
   const [lastNameError, setLastNameError] = useState<string | undefined>(
     undefined
   );
+  const [firstNameError, setFirstNameError] = useState("");
 
   const isPhoneNumberAllNumbers = /^[0-9]*$/.test(phone);
   const phoneValidator = /^05(?!6)\d{8}$/;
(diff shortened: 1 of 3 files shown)


packages/donor/src/screens/extendedSignup/PersonalDetails.tsx
```diff
@@ -55,15 +57,34 @@ export default function PersonalDetails(props: PersonalDetailsProps) {
...
+  const areAllFieldsValid =
+    !firstNameError && !lastNameError && isValidPhone && bloodType;
+  const allFieldsHaveValue =
+    !isEmpty(firstName) && !isEmpty(lastName) && !isEmpty(phone);
+
+  const setFirstNameAndValidate = (newFirstName: string) => {
+    let firstNameErrorMessage = "";
+
+    if (newFirstName.length < 2) {
+      firstNameErrorMessage = "נא להזין שם פרטי";
```
yaron1m, line 69: I think the message should be something like ״שם אינו תקין״

What do you think?

packages/donor/src/screens/extendedSignup/PersonalDetails.tsx
```diff
@@ -129,7 +155,7 @@ export default function PersonalDetails(props: PersonalDetailsProps) {
         <Button
           onClick={onSave}
           title={props.buttonText}
-          isDisabled={!areAllFieldsValid}
+          isDisabled={!areAllFieldsValid || !allFieldsHaveValue}
```
yaron1m, line 158: This PR looks great, and I looked at the Storybook of it, and saw the error shows when I type the first letter, and disappears when I have more than 2, same for the phone number.

I have a suggestion on how to make the experience better:
1. Always enable the save button.
2. On save - check all values
3. If valid - proceed.
4. If not - show errors when needed and don't proceed.

What do you think?

packages/donor/src/screens/extendedSignup/PersonalDetails.tsx
ritabiniash: Sure :)

packages/donor/src/screens/extendedSignup/PersonalDetails.tsx
ritabiniash: Emmm personally I don't like forms with that behavior because you to go backwards and fill it again and it feels like you are filling it twice, on the other hand it is a very short form. Are you aware of any plans to expand it ? @yaron1m

yaron1m: We do plan to add a field or 2 in the future.

Anyway - I trust your judgment. If you think it is a better experience this way - I'm in

yaron1m (approved):

PR #314, opened 2022-01-02
changing login page

yedidya03:
image

image


packages/coordinator/src/screens/signIn/SignInScreen.tsx
```diff
@@ -34,11 +40,31 @@ export default function SignInScreen(props: SignInScreenProps) {
     }
   };
 
+  const resetPassword = async () => {
```
yaron1m, line 43: Why did you decide to merge the two screens into one?

Remember that there's also a popup to add here after the email is sent. I am not sure it's a good idea to add it to this screen as well

yedidya03: It is pretty much the same screen, I didn't want to duplicate it. An other approach could be to make a different component of all the screen structure and use it in each screen, but it look quiet the same to me..

Another benefit of doing it in the same screen is that you don't need to type the email again when you forget the password.

yaron1m: What about the missing popup?

image

Anyway, I understand the screens may be very similar, but they may diverge in the future. Sign in screen may have a button to register, Reset password screen can have a "contact help" button.

You could put them in the same folder and share the CSS file if it helps, although I would try to avoid that as well.

packages/coordinator/src/assets/icons/chevron-right-small.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="8" height="12" viewBox="0 0 8 12" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M1 11L6.38462 6L1 1" stroke="#474747" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
```
yaron1m, line 2: Didn't I already add this icon?

packages/coordinator/src/screens/AuthScreens/ResetPasswordScreen.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { TestUtils } from "@zm-blood-components/common";
+import ResetPasswordScreen, {
+  ResetPasswordScreenProps,
+} from "./ResetPasswordScreen";
+
+export default {
+  component: ResetPasswordScreen,
+  title: "Screens/Reset Password Screen",
+  parameters: { layout: "fullscreen" },
+};
```
yaron1m, line 11: add `as Meta` (see in other stories)

yaron1m (approved):

packages/coordinator/src/assets/icons/chevron-right-small.svg
yedidya03: yes, for the back button

packages/coordinator/src/screens/AuthScreens/ResetPasswordScreen.stories.tsx
yedidya03: done

PR #315, opened 2022-01-03
New Coordinator header variations

yaron1m:
image


yaron1m:
@yedidya03 The build breaks because I didn't want to conflict with your sign in screen changes, please submit it when you can

PR #316, opened 2022-01-03
Show popup when booking an appointment for the same day

yaron1m:
image


packages/donor/src/components/basic/Popup/Popup.tsx
```diff
@@ -50,7 +52,10 @@ export default function Popup({
         {image && <img src={image} alt={"popup"} className={styles.image} />}
         <div className={styles.popupText}>
           <div className={styles.popupTextTitle}>{title}</div>
-          <div className={styles.popupTextContent}>{content}</div>
+          <div className={styles.popupTextContent}>
+            {content}
+            {children}
+          </div>
```
yedidya03, line 58: you can remove the content prop if you use children

packages/donor/src/screens/UpcommingDonation/SameDayDonationUtil.ts
```diff
@@ -0,0 +1,19 @@
...
+  const donationStartTime = new Date(bookedAppointment.donationStartTimeMillis);
+  const bookingTime = new Date(bookedAppointment.bookingTimeMillis);
+  if (!DateUtils.isSameDay(now, donationStartTime)) {
+    return false;
+  }
+
+  // If appointment was booked more than 10s ago
+  if (now.getTime() - bookingTime.getTime() > 10_000) {
+    return false;
+  }
```
yedidya03, line 16: what is that for?

yaron1m: Without this, a user could schedule on Sunday an appointment starting on Monday, and then re-open the app on Monday and see the popup.

Here I verify that the appointment was booked very recently

packages/donor/src/components/basic/Popup/Popup.tsx
yaron1m: Done

packages/donor/src/screens/UpcommingDonation/SameDayDonationUtil.ts
yaron1m: Added a comment

yedidya03: but this code is executed only if the appointment is in the same day, so in this usecase the code will not get there at all.

packages/donor/src/components/basic/Popup/Popup.tsx
```diff
@@ -44,13 +46,17 @@ export default function Popup({
     onClose = onBack;
   }
 
+  if (content) {
+    children = content;
+  }
```
yedidya03, line 51: why do you need content at all?

yaron1m: Done

yedidya03 (approved):
Nice Work!

PR #318, opened 2022-01-04
adding all to reports page and funciton

yedidya03:
image

### DIFF
--- a/packages/functions/src/dal/AdminDataAccessLayer.ts
+++ b/packages/functions/src/dal/AdminDataAccessLayer.ts
@@ -97,5 +97,7 @@ export function getValidHospitalsOrThrow(
         : [hospital];
   }
 
-  throw Error("Unfamiliar coordinator role");
+  throw Error(
+    "Error while trying to get coordinator hospital : 'Unfamiliar coordinator role'"
+  );
 }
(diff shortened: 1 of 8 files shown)


packages/functions/src/dal/AdminDataAccessLayer.ts
```diff
@@ -60,3 +65,37 @@ export async function getCoordinatorDonors(
...
+        throw Error("Coordinator has no permissions for hospital");
+      }
+      return [hospital];
+    case CoordinatorRole.GROUP_COORDINATOR:
+      return hospital === HospitalUtils.ALL_HOSPITALS_SELECT
+        ? coordinator.hospitals ?? []
+        : [hospital];
+  }
+
+  throw Error("Unfamiliar coordinator role");
```
yaron1m, line 100: Unfamiliar coordinator hospital

yaron1m (approved):
Please verify the hospital name appears in the exported CSV (when there are multiple hospitals in the response)

packages/functions/src/dal/AdminDataAccessLayer.ts
yedidya03: if the code gets there it means that we have a role that was not covered in the switch case

yaron1m (approved):

PR #323, opened 2022-01-06
Schedule picker

yaron1m:
image

### DIFF
--- a/packages/coordinator/src/components/Schedule/ScheduleWeekPicker/SchedulePicker.tsx
+++ b/packages/coordinator/src/components/Schedule/ScheduleWeekPicker/SchedulePicker.tsx
@@ -1,0 +1,157 @@
+import React, { useState } from "react";
+import styles from "./SchedulePicker.module.scss";
+import { DateUtils, Hospital, LocaleUtils } from "@zm-blood-components/common";
+import { ReactComponent as ChevronLeft } from "../../../assets/icons/chevron-left-small.svg";
+import { ReactComponent as ChevronRight } from "../../../assets/icons/chevron-right-small.svg";
+import { ReactComponent as ChevronDown } from "../../../assets/icons/chevron-down.svg";
+import { ReactComponent as Calendar } from "../../../assets/icons/calendar.svg";
+import { ReactComponent as Selected } from "../../../assets/icons/selected.svg";
+import classNames from "classnames";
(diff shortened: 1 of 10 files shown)


packages/coordinator/src/components/Schedule/ScheduleWeekPicker/SchedulePicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+        <HospitalSelected
+          {...props}
+          showHospitalPicker={showHospitalPicker}
+          setShowHospitalPicker={setShowHospitalPicker}
+        />
+      </div>
+      <HospitalPicker
+        {...props}
+        showHospitalPicker={showHospitalPicker}
+        setShowHospitalPicker={setShowHospitalPicker}
```
yedidya03, line 36: you can pass only the set function and pass to it (lastVal) => { return munipilatedLastVal }.
(I think it is slightly cleaner because there are less parameters to pass - do what you think)

packages/coordinator/src/components/Schedule/ScheduleWeekPicker/SchedulePicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const [showHospitalPicker, setShowHospitalPicker] = useState(false);
+
+  return (
+    <>
+      <div className={styles.container}>
+        <DatePicker {...props} />
+        <HospitalSelected
+          {...props}
+          showHospitalPicker={showHospitalPicker}
+          setShowHospitalPicker={setShowHospitalPicker}
```
yedidya03, line 30: here too

yedidya03 (approved):

packages/coordinator/src/components/Schedule/ScheduleWeekPicker/SchedulePicker.tsx
yaron1m: Notice I need both of them, as I also read the value of showHospitalPicker

packages/coordinator/src/components/Schedule/ScheduleWeekPicker/SchedulePicker.tsx
yaron1m: Same

PR #325, opened 2022-01-07
Unsubscribe page

ritabiniash:
image


packages/donor/src/screens/unsubscribe/UnsubscribedScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+  isLoggedIn: boolean;
+  user?: Donor;
+}
+
+const UnsubscribedScreen = (props: UnsubscribedScreenProps) => {
+  const { user, isLoggedIn } = props;
+  const profileLink = <Link to={MainNavigationKeys.MyProfile}>הפרופיל</Link>;
+  const GOODBYE = "להתראות !";
+  const EMAIL_REMOVED = `${
+    isLoggedIn ? user?.email : "האימייל שלך"
```
yaron1m, line 18: Even if the user is logged in, nothing says it is logged in with the same email we just unsubscribed, so we might be lying to the user here.

I think we should just always say "האימייל שלך". What do you think?

packages/donor/src/utils/StorybookMocks.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  Donor,
+  BloodType,
+  DonorNotificationSettings,
+} from "@zm-blood-components/common";
+
+const notificationSettingsMock: DonorNotificationSettings = {
+  disableEmailNotifications: true,
+};
+export const mockUser: Donor = {
```
yaron1m, line 10: Nice!
Please move the file to `src/__test__`.

Also, I suggest calling the object `sampleUser` because we already have similar thing in `functions`

packages/donor/src/screens/unsubscribe/UnsubscribedScreen.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { UnsubscribedScreenProps } from "./UnsubscribedScreen";
+import { mockUser } from "../../utils/StorybookMocks";
+
+export default {
+  component: UnsubscribedScreen,
+  title: "Screens/Unsubscribed",
+  parameters: { layout: "fullscreen" },
+  decorators: [
+    (Story: any) => (
+      <MemoryRouter>
```
yaron1m, line 13: I added MemoryRouter to `preview.tsx` so it is not needed here. I think you can remove the entire `decorators`

yaron1m:
Please note you have merge conflicts with my Router changes

packages/donor/src/screens/unsubscribe/UnsubscribedScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+import styles from "./UnsubscribedScreen.module.scss";
+
+export interface UnsubscribedScreenProps {
+  isLoggedIn: boolean;
+  user?: Donor;
+}
+
+const UnsubscribedScreen = (props: UnsubscribedScreenProps) => {
+  const { user, isLoggedIn } = props;
+  const profileLink = <Link to={MainNavigationKeys.MyProfile}>הפרופיל</Link>;
```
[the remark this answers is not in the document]
ritabiniash: ok, why do we prefer that ? @yaron1m

packages/donor/src/screens/unsubscribe/UnsubscribedScreen.tsx
ritabiniash: sure, we can add in the task that will redirect to this page to add the email in the redirection, for now I will remove it

packages/donor/src/screens/unsubscribe/UnsubscribedScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const GOODBYE = "להתראות !";
+  const EMAIL_REMOVED = `${
+    isLoggedIn ? user?.email : "האימייל שלך"
+  } הוסר בהצלחה מרשימת התפוצה של בנק מרכיבי הדם`;
+  const REGRET = "התחרטת? תמיד נשמח שוב לראותך!";
```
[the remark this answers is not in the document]
ritabiniash: I think it is much cleaner to understand to component structure. Especially when our text is in Hebrew

packages/donor/src/screens/unsubscribe/UnsubscribedScreen.tsx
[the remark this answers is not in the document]
yaron1m: That way we control how links look like in one place. 

If we someday want to change the link color to be green, we can easily do that

packages/donor/src/screens/unsubscribe/UnsubscribedScreen.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const UnsubscribedScreen = () => {
+  const navigate = useNavigate();
+
+  const profileLink = (
+    <Button
+      onClick={() => navigate(MainNavigationKeys.MyProfile)}
+      title={"הפרופיל"}
+      variant={ButtonVariant.text}
+      className={styles.ProfileLink}
```
ritabiniash, line 16: Added the Button but I think that a common component for link is an overkill because : 
1. Links color (or any other attribute) shouldn't change , it is a global convention
2. Noticed hoe much code I needed to add for that : `useHistory`, `Button` imports, and styling, while with `Link` we only had to add one simple line
@yaron1m WDYT ?

yaron1m: I think there's no one right way.

I understand your arguments, and you are not wrong.

Both ways look OK to me, I just want us to be consistent.

yaron1m (approved):

PR #338, opened 2022-01-19
Function/donor swap appointments

evbambly:
Swap donations by donor Firebase Function


evbambly:
Implementation notes:

* Extracted Validation functions for Book and Cancel functions.
   Note - this means the Swap function runs both validations, 
   as well as both being run upon Book/Cancel.
   This is less performative, and results in a valid test time of 5300~ ms.
   Please advise whether further optimization is required in your opinion. 

* ShouldNotify added as a Boolean, 
   as Cancel/Book notifications should be disabled for Swap, 
   which will notify Coordinators in a unique format

packages/functions/src/coordinator/CoordinatorBookAppointmentHandler.ts
```diff
@@ -19,6 +19,7 @@ export default async function (
   return bookAppointment(
     request.donorId,
     request.appointmentIds,
+    true,
```
yaron1m, line 22: Should this be false?

packages/functions/src/common/CancelAppointmentHelper.ts
```diff
@@ -0,0 +1,69 @@
...
+  shouldNotify: boolean
+) {
+  const donorId = callerId;
+
+  const { isValid, invalidReason, appointment } =
+    await validateCancelAppointment(request.appointmentId, donorId);
+
+  if (!isValid || !appointment) {
+    throw new Error(invalidReason);
+  }
```
yaron1m, line 29: I don't really like this pattern.

What do you think about:

```
const appointment = getAppointment();

validate(appointment); // If the appointment is invalid, this will throw

// ... continue code here
```

Meaning - there's no reason to return a valid response.
If the appointment is not valid, just throw.

packages/functions/src/common/CancelAppointmentHelper.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  if (shouldNotify) {
```
yaron1m, line 31: Notify after you commit the change, i.e. line 46

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+  const cancelAppointmentValidation = validateCancelAppointment(
+    request.cancelAppointmentId,
+    callerId
+  );
+
+  // Unified promises due to low performance
+  const [
+    { status: bookAppointmentStatus },
+    { isValid: isCancelValid, invalidReason: cancellationInvalidReason },
+  ] = await Promise.all([
```
yaron1m, line 30: If we ever decide to validate the donor doesn't already have a pending appointment, this will fail.

I think we should first cancel the appointment, and only after that book the other one

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+): Promise<FunctionsApi.SwapAppointmentResponse> {
+  const bookAppointmentValidation = validateBookAppointment(
+    request.bookAppointmentIds,
+    callerId
+  );
+
+  const cancelAppointmentValidation = validateCancelAppointment(
+    request.cancelAppointmentId,
+    callerId
+  );
```
yaron1m, line 24: Do both of these fetch the appointment?

The should not. They should get it as an argument

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
yaron1m: The low performance is probably not due to this, but due to the cold start of the functions.

If you run this 3 times in a row, how long does it take to run?

packages/functions/src/donor/CancelAppointmentHandler.ts
```diff
@@ -1,41 +1,9 @@
...
-  const donor = await getDonor(donorId);
-  notifyOnCancelAppointment(appointment, donor!).catch((e) =>
-    console.error("Error notifying on cancelled appointment", appointment.id, e)
-  );
-
-  const updatedAppointment =
-    DbAppointmentUtils.removeDonorFromDbAppointment(appointment);
-
-  await setAppointment(updatedAppointment);
+  await cancelAppointment(callerId, request, true);
```
yaron1m, line 8: Instead of extracting this to a file and call it with true or false for notifying, I suggest something else:

1. Fetch appointment
2. Extract validation logic to util (like you already did, with small changes)
3. Remove donor and save appointment.
4. Notify user.

Then, in the other function, just run steps 1-3

yaron1m (commented):
Apologies for the late review

packages/functions/src/coordinator/CoordinatorBookAppointmentHandler.ts
evbambly: Coordinator appointments currently trigger a notification.
Should I disable it?

Should we have a centralized NotificationHandler to receive relevant information 
and a NotificationType enum, instead of a Boolean?

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
evbambly: 1. Should we not change the structure of the function then?
2. Currently I run both validations simultaneously and then book and then cancel, 
with an internal validation for each.
This means running 4 validations. 
If we should validate + cancel and then validate + book, why not skip the external validation altogether?
3. Running the validations simultaneously saved 600ms over synchronously,
running the tests repeatedly did not yield significant improvements to performance.

evbambly: Function now Cancels, then books the appointments.

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
evbambly: See above comment

evbambly: The extracted validation is diluted that way,
isn't there another option?

packages/functions/src/donor/CancelAppointmentHandler.ts
evbambly: Seems undry

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
evbambly: > If we ever decide to validate the donor doesn't already have a pending appointment, this will fail.

Then we can extract that booking status from the book validation function and uniquely pass it as acceptable,
only in the case of SWAP.

packages/functions/src/coordinator/CoordinatorBookAppointmentHandler.ts
evbambly: BookAppointment is sufficiently complex in my opinion (70 rows),
to not write twice.
CancelAppointment was removed.

packages/functions/src/common/BookAppointmentHelper.ts
```diff
@@ -14,73 +13,67 @@ import {
...
+    donorDetails
   );
-  if (availableAppointments.length === 0) {
-    // None of the requested appointments is available
-    return {
-      status: FunctionsApi.BookAppointmentStatus.NO_AVAILABLE_APPOINTMENTS,
-    };
+
+  if (
+    !appointmentToBook ||
```
yaron1m, line 46: I think this is redundant. 
checking the status is enough

packages/functions/src/common/BookAppointmentHelper.ts
```diff
@@ -103,3 +96,40 @@ function updateDonorAsync(donor: DbDonor, hospital: Hospital) {
...
   return setDonor(updatedDonor);
 }
+
+export function validateBookAppointment(
+  appointments: DbAppointment[],
+  donorId: string,
+  coordinatorId?: string,
+  donorDetails?: MinimalDonorDetailsForAppointment
+): ValidBookAppointmentResponse {
+  if (!appointments || appointments.length === 0) {
```
yaron1m, line 106: Why do you check `!appointments`? It can't be undefined here (or the build would fail)

packages/functions/src/common/BookAppointmentHelper.ts
```diff
@@ -14,73 +13,67 @@ import {
...
-    );
+    if (shouldNotify) {
+      notifyOnAppointmentBooked(appointmentToBook, donor).catch((e) =>
+        console.error(
+          "Error notifying on booked appointment",
+          appointmentToBook.id,
+          e
+        )
+      );
+    }
```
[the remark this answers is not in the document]
evbambly: This can be done later

packages/functions/src/common/BookAppointmentHelper.ts
evbambly: Done

packages/functions/src/common/BookAppointmentHelper.ts
```diff
@@ -15,11 +15,19 @@ import {
...
 import { DbDonor, DbAppointment } from "../function-types";
 import * as DbAppointmentUtils from "../utils/DbAppointmentUtils";
+import { BookAppointmentStatus } from "common/src/functions-api";
 
-export interface ValidBookAppointmentResponse {
+export type ValidBookAppointmentResponse =
+  | ValidBookAppointment
+  | InvalidBookAppointment;
+
+export type ValidBookAppointment = {
```
yaron1m, line 24: I don't think you need to export this one (and the next one)

evbambly: Done

yaron1m: I see you changed only one

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+  getAppointmentsByIds,
+  setAppointment,
+} from "../dal/AppointmentDataAccessLayer";
+import { removeDonorFromDbAppointment } from "../utils/DbAppointmentUtils";
+
+export default async function (
+  request: FunctionsApi.SwapAppointmentRequest,
+  callerId: string
+): Promise<FunctionsApi.SwapAppointmentResponse> {
+  if (!request.cancelAppointmentId) {
```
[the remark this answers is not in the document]
evbambly: Done

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+    throw new Error("No appointment to cancel");
+  }
+
+  const appointmentsToValidate = await getAppointmentsByIds([
+    ...request.bookAppointmentIds,
+    request.cancelAppointmentId,
+  ]);
+
+  const appointmentToCancel = appointmentsToValidate.find(
```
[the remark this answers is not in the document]
evbambly: Done

packages/functions/src/donor/DonorSwapAppointmentHandler.test.ts
```diff
@@ -0,0 +1,272 @@
...
+const SWAP_TEST_HOSPITAL = Hospital.RAMBAM;
+const CREATOR_USER_ID = "creatorUserId";
+const DONOR_ID = "SwapAppointmentHandlerDonorId1";
+const OTHER_DONOR_ID = "SwapAppointmentHandlerDonorId2";
+const APPOINTMENT_TO_BOOK_1 = "SwapAppointmentHandlerAppointment1";
+const APPOINTMENT_TO_BOOK_2 = "SwapAppointmentHandlerAppointment2";
+const OTHER_DONATION_OF_USER = "SwapAppointmentHandlerAppointment3";
+const APPOINTMENT_TO_CANCEL = "SwapAppointmentHandlerAppointment4";
+
+beforeAll(async () => {
```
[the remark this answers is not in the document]
evbambly: Done

packages/functions/src/donor/DonorSwapAppointmentHandler.test.ts
```diff
@@ -0,0 +1,272 @@
...
+
+afterEach(async () => {
+  await DonorDataAccessLayer.deleteDonor(DONOR_ID);
+  await DonorDataAccessLayer.deleteDonor(OTHER_DONOR_ID);
+  await deleteAppointmentsByIds([
+    APPOINTMENT_TO_BOOK_1,
+    APPOINTMENT_TO_BOOK_2,
+    OTHER_DONATION_OF_USER,
+    APPOINTMENT_TO_CANCEL,
+  ]);
```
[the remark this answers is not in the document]
evbambly: Done

packages/functions/src/donor/DonorSwapAppointmentHandler.test.ts
```diff
@@ -0,0 +1,272 @@
...
+  getAppointmentsByIds,
+  setAppointment,
+} from "../dal/AppointmentDataAccessLayer";
+import { expectAsyncThrows } from "../testUtils/TestUtils";
+import * as admin from "firebase-admin";
+import { sampleUser } from "../testUtils/TestSamples";
+import { BookAppointmentStatus } from "@zm-blood-components/common/src/functions-api";
+import { AppointmentStatus } from "@zm-blood-components/common/src";
+import { DbAppointment, DbDonor } from "../function-types";
+import { SwapAppointmentRequest } from "@zm-blood-components/common/lib/functions-api";
```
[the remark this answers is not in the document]
evbambly: Done

packages/functions/src/donor/DonorSwapAppointmentHandler.test.ts
```diff
@@ -0,0 +1,272 @@
...
+  await saveAppointment(APPOINTMENT_TO_BOOK_2, 8, OTHER_DONOR_ID);
+
+  const response = await wrapped(
+    swapAppointmentRequest([APPOINTMENT_TO_BOOK_1, APPOINTMENT_TO_BOOK_2]),
+    {
+      auth: {
+        uid: DONOR_ID,
+      },
+    }
+  );
```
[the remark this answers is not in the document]
evbambly: Done

packages/functions/src/donor/CancelAppointmentHandler.ts
```diff
@@ -17,25 +18,22 @@ export default async function (
     throw new Error("No appointment to cancel");
   }
 
-  const appointmentToCancel = await getAppointmentsByIds([
-    request.appointmentId,
-  ]);
-  if (appointmentToCancel.length !== 1) {
+  const foundAppointments = await getAppointmentsByIds([request.appointmentId]);
```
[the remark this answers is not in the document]
evbambly: In a separate PR

packages/functions/src/common/BookAppointmentHelper.ts
evbambly: The ValidBookAppointment is needed as an import for the test file

yaron1m (approved):

packages/functions/src/donor/DonorSwapAppointmentHandler.ts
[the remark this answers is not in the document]
yaron1m: I meant after the `find`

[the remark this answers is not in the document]
yaron1m: Nevermind

yaron1m:
Great work!

PR #341, opened 2022-01-25
added padding to pop up bottom

ritabiniash:
image

### DIFF
--- a/packages/coordinator/src/components/Popup/Popup.tsx
+++ b/packages/coordinator/src/components/Popup/Popup.tsx
@@ -1,4 +1,5 @@
 import { noop } from "lodash";
+import clsx from "clsx";
 import {
   Dialog,
   DialogActions,
@@ -43,6 +44,9 @@ export function Popup(props: PopupProps) {
       textAlign: "center",
       paddingBottom: "10px",
     },
+    bottom: {
+      paddingBottom: "30px",
+    },
   });
 
   const classes = centerStyle();
@@ -51,7 +55,7 @@ export function Popup(props: PopupProps) {
     <Dialog open={open} onClose={onClose} fullWidth={true}>
       <DialogTitle>{title}</DialogTitle>
       <DialogContent classes={classes}>{children}</DialogContent>
-      <DialogActions classes={classes}>
+      <DialogActions className={clsx(classes.root, classes.bottom)}>
         {primaryButtonText && (
           <Button
             title={primaryButtonText}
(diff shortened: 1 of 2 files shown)


packages/coordinator/src/components/Popup/Popup.tsx
```diff
@@ -51,7 +55,7 @@ export function Popup(props: PopupProps) {
     <Dialog open={open} onClose={onClose} fullWidth={true}>
       <DialogTitle>{title}</DialogTitle>
       <DialogContent classes={classes}>{children}</DialogContent>
-      <DialogActions classes={classes}>
+      <DialogActions className={clsx(classes.root, classes.bottom)}>
```
yaron1m, line 58: We don't have a dependency on `clsx`.

Please use `classNames(...)`

packages/coordinator/src/components/Popup/Popup.tsx
```diff
@@ -43,6 +44,9 @@ export function Popup(props: PopupProps) {
       textAlign: "center",
       paddingBottom: "10px",
     },
+    bottom: {
+      paddingBottom: "30px",
```
yaron1m, line 48: Do we want this in all popups? Or just the delete appointment one?

ritabiniash: all pf them

packages/coordinator/src/components/Popup/Popup.tsx
ritabiniash: ok will change. it worked locally :O

yaron1m: I guess we have it as a transitive dependency

packages/coordinator/src/components/Popup/Popup.tsx
yaron1m: Did you check how the other popups look after this change? Do they look OK?

ritabiniash: Good point :) yes checked them 
BTW we wanted this behavior when we originally wrote the pop-up but there was an issue so it was reverted

yaron1m: Great.

Yedidya just pushed another usage of the popup, please take a look at it as well, and if it looks good let's merge this :)

yaron1m (approved):

PR #342, opened 2022-01-29
Add 'lastDonationTime' to donor

ritabiniash:
A draft because the implementation might change, but i want to make sure I got the BE right :)

Open questions
1. As i understand, we need to update this field when an appointment is marked is completed but maybe that's wrong
2. Haven't tested it locally 
3. Need to add donor id validation 
4. Need to add tests


packages/functions/src/donor/CompleteAppointmentHandler.test.ts
```diff
@@ -136,6 +136,8 @@ test("Valid request when caller is a coordinator", async () => {
     },
   });
 
+  // Add donor test
```
yaron1m, line 139: Please add a test :)

packages/functions/src/function-types.ts
```diff
@@ -24,6 +24,9 @@ export type DbDonor = {
   lastBookingTime?: firestore.Timestamp;
   lastBookedHospital?: Hospital; // The hospital of the last booked donation of the donor
 
+  // Fields changing after completing a donation
+  lastDonationTime?: firestore.Timestamp;
```
yaron1m, line 28: What do you think about `lastCompletedDonationTime`?

packages/functions/src/donor/CompleteAppointmentHandler.ts
```diff
@@ -68,6 +70,11 @@ export async function completeAppointmentFunc(
 
   await AppointmentDataAccessLayer.setAppointment(updatedAppointment);
 
+  if (appointment.status == AppointmentStatus.COMPLETED) {
+    // TODO : verify if donorId is valid, should be done here or in DAL ?
```
yaron1m, line 74: I think you should validate here.

BTW, if donorId is MANUAL_DONOR_ID I think you should not run this update

PR #343, opened 2022-01-31
added ios calender event

shwinky:


shwinky:
@yaron1m should we add yarn.lock to gitignore?

yaron1m:
> @yaron1m should we add yarn.lock to gitignore?

The file is important, it specifies the dependencies used and their specific versions. 

Is there a problem with it?

packages/donor/src/screens/UpcommingDonation/EventCalendarLink.tsx
```diff
@@ -62,17 +61,42 @@ export default function EventCalendarLink({
...
+    const filename = "download.ics";
+    const blob = new Blob([calendarUrl], {
+      type: "text/calendar;charset=utf-8",
+    });
+    const link = document.createElement("a");
+    link.href = window.URL.createObjectURL(blob);
+    link.setAttribute("download", filename);
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
```
yaron1m, line 88: Why do you need to create an `a` element here, and not in line 99?

Would this work?

```
        <div className={styles.link} >
               <a href={...} download={...}>הוספה ליומן</a>
        </div>
```

shwinky: fixed

shwinky:
@yaron1m yarn.lock gets overrided when you yarn install... if the server does yarn install before it runs the code then it has no effect on it....you can use the yark.lock to troubleshoot issues if you want to use the same version you had on your local computer.....but i don't think it should be updated everytime one of use addes a new package and installs

yaron1m:
I still think we need this file.

It is not overridden every time you install, only when something changes.
The lock file let's you be explicit about dependencies version, and I think this is a good thing.

Why do you think we should remove it?

packages/donor/src/screens/UpcommingDonation/EventCalendarLink.tsx
```diff
@@ -62,17 +53,37 @@ export default function EventCalendarLink({
...
         <div className={styles.link} onClick={addToGoogleCalendar}>
           הוספה ליומן
         </div>
       ) : (
-        <div className={styles.link}>
-          <ICalendarLink event={getAppleCalendarEvent()}>
-            הוספה ליומן
-          </ICalendarLink>
+        <div className={styles.link} onClick={addToIOSCalendar}>
+          הוספה ליומן
```
yaron1m, line 86: These 2 divs are the same, they only have different onClicks.

I suggest you merge them, and create another onClick function that will call the relevant `addTo...` based on the device

yaron1m (approved):

PR #348, opened 2022-02-07
Util/get appointment by

evbambly:
getAppointmentById added, utilizing the original getAppointmentsByIds.

getAppointmentByIdOrThrow is the safe version, 
throwing instead of returning undefined, to standardize the resulting error.

### DIFF
--- a/packages/functions/src/dal/AppointmentDataAccessLayer.ts
+++ b/packages/functions/src/dal/AppointmentDataAccessLayer.ts
@@ -33,7 +33,7 @@ export async function getAppointmentByIdOrThrow(
   const appointment = foundAppointments[0];
 
   if (!appointment) {
-    throw new Error("Appointment not found");
+    throw new Error(`Appointment not found. Id ${appointmentId}`);
   }
 
   return appointment;
--- a/packages/functions/src/dal/AppointmentDataAccessLayer.ts
+++ b/packages/functions/src/dal/AppointmentDataAccessLayer.ts
@@ -26,6 +26,25 @@ export async function getAppointmentsByIds(
   return _.flatMap(snapshots, (s) => toDbAppointments(s));
 }
 
+export async function getAppointmentById(
+  appointmentId: string
+): Promise<DbAppointment | undefined> {
+  const foundAppointments = await getAppointmentsByIds([appointmentId]);
+  return foundAppointments[0];
+}
+
+export async function getAppointmentByIdOrThrow(
+  appointmentId: string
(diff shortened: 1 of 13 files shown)


packages/functions/src/utils/DbAppointmentUtils.ts
```diff
@@ -120,6 +121,25 @@ export function toBookedAppointmentSync(
   }
 }
 
+export async function getAppointmentById(
```
yaron1m, line 124: These 2 functions should be in the Appointment DAL, not here

packages/functions/src/utils/DbAppointmentUtils.ts
```diff
@@ -120,6 +121,25 @@ export function toBookedAppointmentSync(
...
+  appointmentId: string
+): Promise<DbAppointment> {
```
yaron1m, line 126: If this doesn't throw, it should return `DbAppointment | undefined`

packages/functions/src/utils/DbAppointmentUtils.ts
evbambly: Done

packages/functions/src/utils/DbAppointmentUtils.ts
evbambly: Done

evbambly: None of the functions are built to handle undefined,
I removed the unsafe method

yaron1m: Great

packages/functions/src/dal/AppointmentDataAccessLayer.ts
```diff
@@ -26,6 +26,19 @@ export async function getAppointmentsByIds(
...
 }
 
+export async function getAppointmentByIdOrThrow(
+  appointmentId: string
+): Promise<DbAppointment> {
+  const foundAppointments = await getAppointmentsByIds([appointmentId]);
+  const appointment = foundAppointments[0];
+
+  if (!appointment) {
+    throw new Error("Appointment not found");
```
yaron1m, line 36: Consider adding the appointment id to the error message

yaron1m (approved):

packages/functions/src/dal/AppointmentDataAccessLayer.ts
evbambly: Done

PR #351, opened 2022-02-11
Coordinator live updates

yaron1m:


packages/common/src/types.ts
```diff
@@ -117,3 +118,9 @@ export enum AppointmentStatus {
 }
 
 export const MANUAL_DONOR_ID = "manual";
+
+export type CoordinatorUpdate = {
```
ritabiniash, line 122: Maybe change to `CoordinatorUpdateProps` ? The current name implies that this is an action

packages/coordinator/src/store/appointments/actions/CoordinatorUpdateAction.ts
```diff
@@ -0,0 +1,38 @@
+import { CoordinatorUpdate } from "@zm-blood-components/common";
+import { ThunkAction } from "../../store";
```
ritabiniash, line 2: What is `ThunkAction` ?

yaron1m: Redux Thunk is a middleware for redux that lets you dispatch other actions and access the app state when running an action: https://github.com/reduxjs/redux-thunk

What I import here is the typing of thunk actions for our app (i.e. thunk + the structure of our state with appointments and stuff)

packages/common/src/types.ts
yaron1m: I think `props` implies this is an object passed to a component.

How about `CoordinatorLastChangeDetails`?

packages/functions/src/coordinator/DeleteAppointmentsHandler.ts
```diff
@@ -37,6 +38,8 @@ export default async function (
     await AppointmentDataAccessLayer.deleteAppointmentsByIds([appointmentId]);
   }
 
+  await setCoordinatorUpdate(appointment.hospital, callerId);
```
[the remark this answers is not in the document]
yaron1m: I thought about making this part of the props of `RequestHandler<T>` but I don't think it makes it much better.

Do you have any suggestions?

PR #353, opened 2022-02-15
Bugfix

evbambly:
About page crashes in all environments.


yaron1m (approved):
Thanks for this

Please also open a ticket in Trello so Nir would know to verify this

evbambly:
https://trello.com/c/WbH0hWmn/503-about-page-crashes-on-opening

PR #356, opened 2022-02-15
Add Status Column to Reports

evbambly:
Added Status to the reports.

I changed the order of the columns to match their order in the screen.

Please advise: Where should the Hospital column be?


yaron1m (approved):
Please change the commit title to be more informative

PR #357, opened 2022-02-16
Do not throw when trying to delete a  deleted appointment

yaron1m:
I see such errors in the logs

### DIFF
--- a/packages/functions/src/coordinator/DeleteAppointmentsHandler.test.ts
+++ b/packages/functions/src/coordinator/DeleteAppointmentsHandler.test.ts
@@ -90,15 +90,21 @@ test("User that does not have the right hospital throws exception", async () => 
   );
 });
 
-test("No such appointment throws exception", async () => {
+test("No such appointment throws exception when only removing donor", async () => {
   await createUser(CoordinatorRole.SYSTEM_USER);
 
-  const action = () => callFunction(APPOINTMENT_ID, false, COORDINATOR_ID);
+  const action = () => callFunction(APPOINTMENT_ID, true, COORDINATOR_ID);
 
   await expectAsyncThrows(
     action,
-    `Appointment not found. Id ${APPOINTMENT_ID}`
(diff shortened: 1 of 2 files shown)


packages/functions/src/coordinator/DeleteAppointmentsHandler.ts
```diff
@@ -17,9 +17,22 @@ export default async function (
...
+  ]);
 
+  if (appointments.length === 0) {
+    if (request.onlyRemoveDonor) {
+      throw new Error(
+        "Cannot remove donor from missing appointment - " +
+          request.appointmentId
+      );
+    }
+    functions.logger.debug("No appointment to delete", request.appointmentId);
```
evbambly, line 31: Why is this error silent?

evbambly (approved):

packages/functions/src/coordinator/DeleteAppointmentsHandler.test.ts
```diff
@@ -90,17 +90,23 @@ test("User that does not have the right hospital throws exception", async () =>
...
     action,
-    `Appointment not found. Id ${APPOINTMENT_ID}`
+    `Cannot remove donor from missing appointment - ${APPOINTMENT_ID}`
   );
 });
 
+test("No such appointment fails silently", async () => {
+  await createUser(CoordinatorRole.SYSTEM_USER);
+
+  await callFunction(APPOINTMENT_ID, false, COORDINATOR_ID);
```
evbambly, line 107: Shouldn't there be an assertion?

yaron1m: I implicitly assert there's no exception throwns

packages/functions/src/coordinator/DeleteAppointmentsHandler.ts
yaron1m: That's the target of this PR, to not throw in such cases (cause it is not actually an error)

PR #368, opened 2022-03-11
Add last completed donation time to donor response

ritabiniash:
To check in the FE if[ enough days has passed between donations](https://trello.com/c/HYCpaIJy/459-donor-alert-pop-up-when-trying-to-book-an-appointment-less-than-a-month-since-previous-appointment), we need to have `lastCompletedDonationTime` data on the appState donor (converted to milli seconds as the rest of the dates in the fe)

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -23,6 +23,7 @@ export type Donor = {
   phone: string;
   bloodType: BloodType;
   notificationSettings: DonorNotificationSettings;
+  lastCompletedDonationTimeMillis?: number;
 };
 
 export type MinimalDonorDetailsForAppointment = {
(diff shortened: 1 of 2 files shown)


ritabiniash:
looks like the functions test is failing because of the changes in `common` 
Do you have an idea for a fix ? Locally the functions test passed @yaron1m

yaron1m:
@ritabiniash I am not sure what the issue is, and I have been trying to solve it for the past 2 hours.

I think it is something with our yarn workspace / firebase functions / webpack configuration that somehow changed.

I will update when I find a solution

yaron1m:
@ritabiniash 
So it only took me about 8 hours to crack this, but I think I fixed the build issues of the functions.
I had to move the entire package to the root folder, so this PR will probably have conflicts.
You can either fix them, or since this is only a small change just re-create it in a new PR

ritabiniash:
> @ritabiniash So it only took me about 8 hours to crack this, but I think I fixed the build issues of the functions. I had to move the entire package to the root folder, so this PR will probably have conflicts. You can either fix them, or since this is only a small change just re-create it in a new PR

PR #371, opened 2022-03-12
Alert pop-up when trying to book an appointment less than a month since previous appointment

ritabiniash:
Ticket : https://trello.com/c/HYCpaIJy/459-donor-alert-pop-up-when-trying-to-book-an-appointment-less-than-a-month-since-previous-appointment

When a donor select a slot for appointment, check that a month passed since previous appointment

image


packages/common/src/DateUtils.ts
```diff
@@ -102,6 +102,19 @@ export function DateWithAddedMinutes(date: Date, minutes: number) {
...
 
+export function getNumberOfDaysBetweenDates(start: Date, end: Date) {
+  // One day in milliseconds
+  const oneDay = 1000 * 60 * 60 * 24;
+
+  // Calculating the time difference between two dates
+  const diffInTime = start.getTime() - end.getTime();
+
+  // Calculating the no. of days between two dates
+  const diffInDays = Math.round(diffInTime / oneDay);
```
yaron1m, line 113: Maybe also add `absolute` to always make it a positive number?

packages/common/src/DateUtils.ts
```diff
@@ -102,6 +102,19 @@ export function DateWithAddedMinutes(date: Date, minutes: number) {
   return new Date(date.getTime() + minutes * 60000);
 }
 
+export function getNumberOfDaysBetweenDates(start: Date, end: Date) {
```
yaron1m, line 105: Please add tests to this function

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
```diff
@@ -34,22 +42,59 @@ export function BookDonationScreenContainer({
...
-    />
+    <div>
+      <BookDonationScreen
+        availableAppointments={availableAppointmentsStore.availableAppointments}
+        onSlotSelected={onSlotSelected}
+        firstName={appState.donor?.firstName}
+        isFetching={availableAppointmentsStore.isFetching}
+        defaultHospital={""}
+      />
+      <Popup
```
yaron1m, line 81: The popup should be inside the screen, not in the container.
The container should only have logic

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
```diff
@@ -34,22 +42,59 @@ export function BookDonationScreenContainer({
...
 
   const onSlotSelected = (donationSlot: DonationSlotToBook) => {
-    appointmentToBookStore.setAppointmentToBook(donationSlot);
-    if (isLoggedIn) {
-      navigate(MainNavigationKeys.Questionnaire);
+    setSelectedSlot(donationSlot);
+    const currentDonor = appState.donor;
+    const timeDifference = currentDonor?.lastCompletedDonationTimeMillis
+      ? DateUtils.getNumberOfDaysBetweenDates(
+          new Date(donationSlot.donationStartTimeMillis),
```
yaron1m, line 49: Seems like you have the time in millis, why convert it to Date and then to millis again?

ritabiniash: All of the functions in DateUtils are accepting Date as parameters, I wanted the new function to align

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
ritabiniash: Will move the popup but not sure I understood to the fullest "The container should only have logic", when should we add containers and what code should be there ?

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
yaron1m: I see.

Notice some utils accept `Date | number`. If you change your util to do so it would accept both, and then you'd be able to call it here with the millis directly

packages/donor/src/screens/bookDonation/BookDonationScreen.tsx
```diff
@@ -17,13 +17,16 @@ import NoAppointments from "../../assets/images/NO Appointments.svg";
...
+import Popup, { PopupProps } from "../../components/basic/Popup";
+import WarningLogo from "../../assets/images/warning.svg";
 
 export interface BookDonationScreenProps {
   availableAppointments: AvailableAppointment[];
   isFetching: boolean;
   firstName?: string;
   onSlotSelected: (donationSlot: DonationSlotToBook) => void;
   defaultHospital: Hospital | "";
+  onSlotSelectedPopUpProps: Pick<PopupProps, "open" | "onBack" | "onApproved">;
```
yaron1m, line 29: Why did  you decide to use `Pick<PopupProps, "open" | "onBack" | "onApproved">`?

Personally, having to write a function name in a string is a smell to me.

I suggest you just add 3 explicit props to this screen: `tooClosePopupOpen`, `onApproveCloseAppointment` and `onDeclineCloseAppointment`.

packages/donor/src/screens/bookDonation/BookDonationScreen.tsx
```diff
@@ -82,6 +86,20 @@ export default function BookDonationScreen({
...
         )}
       </div>
+
+      <Popup
+        title={"מועד קרוב מידי"}
+        buttonApproveText={"כן, אשר תור"}
+        open={onSlotSelectedPopUpProps.open}
+        goBackText={"התחרטתי, חזרה למסך התורים"}
+        onApproved={onSlotSelectedPopUpProps.onApproved}
+        onBack={onSlotSelectedPopUpProps.onBack}
```
yaron1m, line 96: Should we also populate onClose? Maybe same as onback?

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
```diff
@@ -69,6 +67,15 @@ export function BookDonationScreenContainer({
     }
   };
 
+  const onSlotSelectedPopUpProps = {
+    open: showWarningPopup,
+    onApproved: continueDonation,
+    onBack: () => {
+      setShowWarningPopup(false);
+      navigate(MainNavigationKeys.BookDonation);
```
yaron1m, line 75: Why do we need to navigate? Aren't we already in the booking screen?

packages/donor/src/screens/bookDonation/BookDonationScreen.stories.tsx
```diff
@@ -21,6 +21,11 @@ const props: BookDonationScreenProps = {
   firstName: "משה",
   defaultHospital: "",
   onSlotSelected: action("onSlotSelected"),
+  onSlotSelectedPopUpProps: {
+    open: false,
```
yaron1m, line 25: Consider adding another story, in which the popup is open

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
```diff
@@ -34,22 +40,53 @@ export function BookDonationScreenContainer({
...
-    if (isLoggedIn) {
-      navigate(MainNavigationKeys.Questionnaire);
+    setSelectedSlot(donationSlot);
+    const currentDonor = appState.donor;
+    const timeDifference = currentDonor?.lastCompletedDonationTimeMillis
+      ? DateUtils.getNumberOfDaysBetweenDates(
+          donationSlot.donationStartTimeMillis,
+          currentDonor.lastCompletedDonationTimeMillis
+        )
+      : MIN_DAYS_BETWEEN_DONATIONS + 1;
```
yaron1m, line 50: IIUC, if the donor has no last donation, you set the last donation to `MIN_DAYS_BETWEEN_DONATIONS + 1` so that the if will fail later.

Let me suggest a cleaner approach:
```
const isAppointmentTooCloseToLastDonation = currentDonor.lastCompletedDonationTimeMillis && DateUtils.getNumberOfDaysBetweenDates(
          donationSlot.donationStartTimeMillis,
          currentDonor.lastCompletedDonationTimeMillis
        ) < MIN_DAYS_BETWEEN_DONATIONS
``` 

Then use this boolean in the if below

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
```diff
@@ -34,22 +40,53 @@ export function BookDonationScreenContainer({
...
+    if (timeDifference < MIN_DAYS_BETWEEN_DONATIONS) {
+      setShowWarningPopup(true);
     } else {
-      navigate(MainNavigationKeys.Register);
+      continueDonation();
     }
   };
 
+  const continueDonation = () => {
+    if (selectedSlot) {
```
yaron1m, line 60: Could `selectedSlot` be undefined?
I think we should throw in such cases (as this should never happen).

What do you think?

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
```diff
@@ -34,22 +40,53 @@ export function BookDonationScreenContainer({
...
+      appointmentToBookStore.setAppointmentToBook(selectedSlot);
+      if (isLoggedIn) {
+        navigate(MainNavigationKeys.Questionnaire);
+      } else {
+        navigate(MainNavigationKeys.Register);
```
yaron1m, line 65: I think we should put the login check first. If not logged in then move to Register.

Only then check if there's a selected slot.

yaron1m (commented):
General flow that we need to check:
1. Non-logged-in user chooses an appointment.
2. Move to login screen.
3. Log in with user with a donation completed 2 days ago.

Will we see the popup or go straight to the questionnaire?
What is the expected behavior in such cases?

yaron1m:
Any update here? @ritabiniash

packages/common/src/DateUtils.test.ts
```diff
@@ -67,4 +68,18 @@ describe("DateUtils", function () {
       new Date(2022, 0, 2, 11, 0)
     );
   });
+
+  test("getNumberOfDaysBetweenDates works", () => {
+    const DAYS_DIFFERENCE = 10;
+    const now = new Date();
+    const yesterday = new Date();
```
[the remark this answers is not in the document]
ritabiniash: This test checks time differences and uses a `const`, there is no dependency in the dates

packages/donor/src/screens/bookDonation/BookDonationScreen.tsx
ritabiniash: I uses it because :

1. To show that there is a strong connection to the popup props
2. Avoid duplication
3. Avoid load on p the props

I don't like the way that pick uses strings too, but using explicit props is a problem if this page will have another popup on the future.

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
ritabiniash: No, the popup should pop before the questionnaire

ritabiniash:
@yaron1m This solution has a bug the prevents scrolling in the appointments page, not sure yet what is the source of the bug but closing this PR for now

packages/common/src/DateUtils.test.ts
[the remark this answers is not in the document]
yaron1m: The dependency here is on the time the tests runs. 
The test is not identical when run at 13:00 or at 13:01 (although right now it doesn't matter).
Still, unit tests should be deterministic and run the same way every time

PR #378, opened 2022-04-12
30 days pop up

ritabiniash:
image

Flow 
If user logged in : 
1. Choose slot
2. Check if 30 days passed donor's last donation
3. if not - the popup will pop 
4. The donor can decide if to continue with donation (goes to Questionnaire) OR to cancel and stay in the book donation screen

if not logged in : when choosing a slot will be redirected to login (usual behavior)


packages/common/src/DateUtils.test.ts
```diff
@@ -67,4 +68,19 @@ describe("DateUtils", function () {
       new Date(2022, 0, 2, 11, 0)
     );
   });
+
+  test("getNumberOfDaysBetweenDates works", () => {
+    const DAYS_DIFFERENCE = 10;
+    const now = new Date();
```
yaron1m, line 74: Please use a const date.
i.e. something like `now = new Date(1649927794000)`.

This makes the tests deterministic

packages/common/src/DateUtils.ts
```diff
@@ -102,6 +102,25 @@ export function DateWithAddedMinutes(date: Date, minutes: number) {
...
+  const date2 = new Date(end);
+
+  // One day in milliseconds
+  const oneDay = 1000 * 60 * 60 * 24;
+
+  // Calculating the time difference between two dates
+  const diffInTime = date1.getTime() - date2.getTime();
+
+  // Calculating the no. of days between two dates
+  const diffInDays = Math.round(diffInTime / oneDay);
```
yaron1m, line 119: Should we use `Math.ceil()`?
What would `ceil` or `round` return when the difference is a bit less than a full day? What is the expected behaviour?

packages/donor/src/screens/bookDonation/BookDonationScreen.stories.tsx
```diff
@@ -116,3 +121,17 @@ Fetching.args = {
...
+
+export const SlotTooCloseDoDonorLastAppointmentPopup = (
+  args: BookDonationScreenProps
+) => <BookDonationScreen {...args} />;
+
+SlotTooCloseDoDonorLastAppointmentPopup.args = {
+  ...props,
+  availableAppointments: sampleAppointments,
+  tooCloseDonationPopupProps: {
+    open: false,
```
yaron1m, line 133: Should this be `true`?

packages/donor/src/screens/bookDonation/BookDonationScreen.stories.tsx
```diff
@@ -116,3 +121,17 @@ Fetching.args = {
   ...props,
   isFetching: true,
 };
+
+export const SlotTooCloseDoDonorLastAppointmentPopup = (
+  args: BookDonationScreenProps
+) => <BookDonationScreen {...args} />;
```
yaron1m, line 127: This should be
`export const SlotTooCloseToDonorLastAppointmentPopup = Template.bind({});`

1. Please fix this.
2. Please adjust the other stories as well.
3. Notice the typo in the story name (do => to)

packages/donor/src/screens/bookDonation/BookDonationScreen.tsx
```diff
@@ -74,6 +81,19 @@ export default function BookDonationScreen({
...
+        title={"מועד קרוב מידי"}
+        buttonApproveText={"כן, אשר תור"}
+        open={tooCloseDonationPopupProps.open}
+        goBackText={"התחרטתי, חזרה למסך התורים"}
+        onApproved={tooCloseDonationPopupProps.onApproved}
+        onBack={tooCloseDonationPopupProps.onBack}
+        image={WarningLogo}
+      >
+        טרם חלפו חודש ימים מיום תרומתך האחרון. האם את/ה בטוח/ה שברצונך לקבוע תור
+        זה ?
```
yaron1m, line 94: Remove space before `?`

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
```diff
@@ -29,26 +31,47 @@ export function BookDonationScreenContainer({
...
   };
 
+  const onSlotSelectedPopUpProps = {
+    open: showWarningPopup,
+    onApproved: () => navigate(MainNavigationKeys.Questionnaire),
+    onBack: () => {
+      setShowWarningPopup(false);
+      navigate(MainNavigationKeys.BookDonation);
+    },
+    onBClose: () => {
```
yaron1m, line 61: onClose

packages/donor/src/state/AppointmentToBookStore.ts
```diff
@@ -30,4 +32,19 @@ export class AppointmentToBookStore {
     this.donationStartTimeMillis = slot.donationStartTimeMillis;
     this.appointmentIds = slot.appointmentIds;
   }
+
+  isDonationValidForDonor(donor: Donor): boolean {
```
yaron1m, line 36: Name is too generic.
How about `isAppointmentTooCloseToLastDonation`?

yaron1m: Also, does this currently return true when donation is not valid?

yaron1m (commented):
Please make sure all comments in #371 were addressed as well,
Especially https://github.com/startach/ZM-BloodComponents/pull/371#pullrequestreview-911509286

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
yaron1m: Also, how is this not a build failure?

yaron1m:
Please notice @evbambly made some changes to components like the `Popup`, adding mandatory analytic props.

@evbambly  - Please advise on changes required in this PR (after a rebase on master, of course)

packages/donor/src/state/AppointmentToBookStore.ts
ritabiniash: 1. Like
2. I see what you mean, there us some mismatch between the the names and meanings, will fix

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
ritabiniash: There is no type and "onClose" was not in the pick @yaron1m

yaron1m: I don't understand, how is `onBClose` not a typo?

ritabiniash: `onSlotSelectedPopUpProps` has no type. 
When it is passed to `BookDonationScreen`, `BookDonationScreen` looks for the props that are in the pick, `onClose` is not in the pick and nobody looks at it

packages/common/src/DateUtils.test.ts
ritabiniash: Although I am using a const to test difference ? Why should it not be deterministic that way ?

packages/donor/src/screens/bookDonation/BookDonationScreen.stories.tsx
ritabiniash: image
Cool to know ! Thank you for commenting this I learned something new

ritabiniash: Getting this error :  
```
 ● Storyshots › Screens/Book Donation Screen › Slot Too Close To Donor Last Appointment Popup

    TypeError: parentInstance.children.indexOf is not a function
```

When using the bind, have you encountered this before ? @yaron1m

yaron1m: Can you share the entire file after your change?

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
yaron1m: So should it be in the pick?

packages/common/src/DateUtils.test.ts
yaron1m: I agree that in this specific test this is not super critical, but what I mean here is a general idea in unit tests. 
Tests should behave and run **exactly the same** every time they run, no matter when they run.

Let's say javascript has some weird bug tomorrow that when `new Date()` is called at times the moon is full, it returns undefined. In this case, our unit test would fail once a month, even if we don't change the code it is testing.

I know this is a really weird example, but my point here is that since we call `new Date()` here, this test relies on the time it is run (as `new Date()` returns a different value based on the time it runs).
If you change the call to something like the first comment here, the test would be more deterministic and the run would be more similar every time it runs

packages/common/src/DateUtils.ts
ritabiniash: Actually copied this sample from the web but 
`round` is the right solution here, `ceil` rounds to the closest number (5.7 -> 6, 5.3 -> 5) while `round` rounds to the lower value (5.7 -> 5, 5.3 -> 5). Since we need the number of full days round is the right one

packages/donor/src/screens/bookDonation/BookDonationScreen.stories.tsx
ritabiniash: ok ok found the issue, it is because the pop up needs a DOM to render itself but in storybook snapshots there is no DOM. We don't have any other snapshots for this kind of popups. This popups are business related and need to be tested manually or with Testim.IO WDYT ?

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
ritabiniash: Yes, added it

packages/donor/src/screens/bookDonation/BookDonationScreen.stories.tsx
yaron1m: I understand. In this case let's just skip the story for this case

ritabiniash:
> Please notice @evbambly made some changes to components like the `Popup`, adding mandatory analytic props.
> 
> @evbambly - Please advise on changes required in this PR (after a rebase on master, of course)

I pulled, I think we should announce it somehow (and any other dev related changes) in the group ?

ritabiniash:
Resolved all comments, thanks for the review ! 

Is the prettier in the CI changed ? I ran `yarn run style` locally and everything looks ok @yaron1m

yaron1m:
> Is the prettier in the CI changed ? I ran yarn run style locally and everything looks ok @yaron1m

This sometimes happens when you `yarn install` not from the root folder (not sure why though).
Please try to `yarn install` from the root folder and then run `yarn run style` again

packages/common/src/DateUtils.ts
yaron1m: Awesome!
BTW, the fact we had this conversation here tells me there aren't enough tests for this function.
I suggest you add 2 more functions, one that checks a case where the diff is a bit more than a (number of) full days, and one that is a bit less.

packages/donor/src/screens/bookDonation/BookDonationScreenContainer.tsx
```diff
@@ -43,9 +43,11 @@ export function BookDonationScreenContainer({
     appointmentToBookStore.setAppointmentToBook(donationSlot);
 
     if (isLoggedIn) {
-      appointmentToBookStore.isDonationValidForDonor(appState.donor!)
-        ? navigate(MainNavigationKeys.Questionnaire)
-        : setShowWarningPopup(true);
+      appointmentToBookStore.isAppointmentTooCloseToLastDonation(
+        appState.donor!
+      )
```
yaron1m, line 48: Consider extracting this to a boolean

packages/donor/src/state/AppointmentToBookStore.ts
```diff
@@ -33,7 +33,7 @@ export class AppointmentToBookStore {
     this.appointmentIds = slot.appointmentIds;
   }
 
-  isDonationValidForDonor(donor: Donor): boolean {
+  isAppointmentTooCloseToLastDonation(donor: Donor): boolean {
```
yaron1m, line 36: Not really sure how to do this (and whether this is hard or not) but this seems like a function that could use a test.

If you have the time, please try to add a test to this one

yaron1m (approved):

yaron1m:
@ritabiniash, there are still some unresolved comments, please take a look at them :)

PR #407, opened 2022-09-23
Analytics V1

evbambly:
Analytics upgrade and revision as per review of the first version:

* Major refactor
* Names of events now accurate
* Login event tracked
* "continue" button -> now "book_appointment"
* Screen_view event now has title
* Popup event bug ("false" event triggered on render) fixed


rbsPREET (approved):
Looks good to me, approved

